PR #78, opened 2023-03-03
Move Shiki to worker process

chaance:
We have the same problem that Kent had with Shiki and memory leaks as we were sharing very similar implementations of our markdown processing work. This PR adapts the same solution by moving the leaky stuff over to a worker process. See https://kentcdodds.com/blog/fixing-a-memory-leak-in-a-production-node-js-app


kentcdodds (approved):
Looks very similar to my implementation which has worked great 👍

PR #172, opened 2024-02-02
feat: query and render the changelog

### LINKED ISSUE #165: Render Changelog on the Website
Our [Changelog](https://remix.run/docs/en/main/start/changelog) doc is sad

image

This was a temporary addition, but we would really prefer to render the full doc on our website. Just a matter of fetching the correct file on that specific route instead of what's in the docs.

mcansh:
probably a better alternative approach to this, but it works...

closes #165

Signed-off-by: Logan McAnsh <logan@mcan.sh>

### DIFF
--- a/app/lib/gh-docs/index.ts
+++ b/app/lib/gh-docs/index.ts
@@ -48,8 +48,9 @@ export function getRepoDocsMenu(ref: string, lang: string) {
   return getMenu(REPO, fixupRefName(ref), lang);
 }
 
-export function getRepoDoc(ref: string, slug: string) {
-  return getDoc(REPO, fixupRefName(ref), slug);
+export function getRepoDoc(ref: string, slug: string, dir?: string) {
+  let filepath = dir ? `${dir}/${slug}` : slug;
+  return getDoc(REPO, fixupRefName(ref), filepath);
 }
 
 function fixupRefName(ref: string) {
--- a/app/lib/gh-docs/index.ts
+++ b/app/lib/gh-docs/index.ts
@@ -48,9 +48,8 @@ export function getRepoDocsMenu(ref: string, lang: string) {
   return getMenu(REPO, fixupRefName(ref), lang);
 }
 
-export function getRepoDoc(ref: string, slug: string, dir?: string) {
-  let filepath = dir ? `${dir}/${slug}` : slug;
-  return getDoc(REPO, fixupRefName(ref), filepath);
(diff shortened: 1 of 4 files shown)


app/routes/docs.$lang.$ref.$.tsx
```diff
@@ -29,8 +29,14 @@ export async function loader({ params, request }: LoaderFunctionArgs) {
...
   let pageUrl = url.protocol + "//" + url.host + url.pathname;
   invariant(params.ref, "expected `ref` params");
+  let splat = params["*"];
+  let isChangelog = splat?.endsWith("/changelog");
   try {
-    let doc = await getRepoDoc(params.ref, params["*"] || "index");
+    let doc = await getRepoDoc(
+      params.ref,
+      isChangelog ? "CHANGELOG" : params["*"] || "index",
+      isChangelog ? "" : "docs",
```
mcansh, line 38: this is the area in particular that could probably be improved..

app/lib/gh-docs/index.ts
```diff
@@ -48,8 +48,9 @@ export function getRepoDocsMenu(ref: string, lang: string) {
   return getMenu(REPO, fixupRefName(ref), lang);
 }
 
-export function getRepoDoc(ref: string, slug: string) {
-  return getDoc(REPO, fixupRefName(ref), slug);
+export function getRepoDoc(ref: string, slug: string, dir?: string) {
+  let filepath = dir ? `${dir}/${slug}` : slug;
+  return getDoc(REPO, fixupRefName(ref), filepath);
```
brookslybrand, line 53: Nice! What do you think about removing the 3rd argument and just require passing in the full path to the slug? I think this simplifies the code you called out

```suggestion
export function getRepoDoc(ref: string, slug: string) {
  return getDoc(REPO, fixupRefName(ref), slug);
```

app/routes/docs.$lang.$ref.$.tsx
brookslybrand: If you go with my other comment, it can simplify it a bit, but you still have still have the issue of hardcoding a "magic" route

Personally I'm pretty okay with this for a one off, and if there are more like this we can start figuring out a better system for it.

My thought at first was to create a unique route for it, but I'm realizing that's a lot more setup for what you've already accomplished.

brookslybrand (approved):
This is great, thank you Logan!

I had one suggestion, you're welcome to take it or leave it

PR #174, opened 2024-02-04
Upgrade remark, rehype, and related packages and fix errors

### LINKED ISSUE #155: Upgrade rehype, remark, and related dependencies
The following packages needed to be bumped to the latest major versions:

- rehype-autolink-headings
- rehype-slug
- rehype-stringify
- remark-gfm
- remark-html
- remark-parse
- remark-rehype
- unified
- unist-util-visit

I upgraded them, deployed, and everything seemed to work fine. However, I got some type errors and am not familiar enough with these packages or the setup to quickly fix it. I figured this might be an issue someone else would be eager to dig their hands into

takagimeow:
Fixes #155 

As mentioned by @brookslybrand , updating packages related to `remark` and `rehype` will result in several type errors.

In this PR, I updated the versions of the packages mentioned in this [issue](https://github.com/remix-run/remix-website/issues/155) to the latest versions and fixed the type errors that occurred.

Updating the following packages:

- rehype-autolink-headings
- rehype-slug
- rehype-stringify
- remark-gfm
- remark-html
- remark-parse
- remark-rehype
- unified
- unist-util-visit

The type errors mainly occurred within two files:

- **/app/lib/mdtut.server.ts**
- **/app/lib/md.server.ts**

## mdtut.server.ts

Due to the removal of `CurrentTree` and the addition of `HeadTree` and `TailTree` to the [`Processor`'s type arguments](https://github.com/unifiedjs/unified/blob/b69689bba52a918d87aa62f295ccffa8d9aa8ef8/lib/index.js#L365-L378), I passed `Hast.Node<Unist.Data>` to each of them.

Also, when passing `Hast.Node<Unist.Data>` as the `CompileTree`, an error occurred stating `Property 'children' is missing in type 'Node<Data>' but required in type 'Root'.`

In the [previous version](https://github.com/unifiedjs/unified/blob/b9358691047ad72e6390dfc3c3eedc15509e01fa/index.d.ts#L84-L85), the description for `CompileTree` was `The node that the compiler receives (and 'run' yields).`, but in the [new version](https://github.com/unifiedjs/unified/blob/b69689bba52a918d87aa62f295ccffa8d9aa8ef8/lib/index.js#L284-L285), it is `Input of 'stringify' (optional).`.

This means that the expected first argument for `processor.stringify()` is not `Node` but `Mdast.Root`.

However, since `TProcessor` is an abstract type, instead of passing specific `Mdast.Root` or `Hast.Root`, I passed `UnistParent`, which is their supertype, as the `CompileTree`.

There were some type errors as well, but these were resolved with type assertions.

## md.server.ts

A type error was encountered with the `Input` type argument of `InternalPlugin`, suggesting that `This type parameter might need an 'extends string | Node | undefined' constraint.`

This error was resolved by explicitly extending `string | Unist.Node | undefined`.

### DIFF
--- a/app/lib/md.server.ts
+++ b/app/lib/md.server.ts
@@ -63,7 +63,7 @@ export async function getProcessor(options?: ProcessorOptions) {
     .use(rehypeAutolinkHeadings);
 }
 
-type InternalPlugin<Input, Output> = Unified.Plugin<
+type InternalPlugin<Input extends string | Unist.Node | undefined, Output> = Unified.Plugin<
   [ProcessorOptions?],
   Input,
   Output
--- a/app/lib/mdtut.server.ts
+++ b/app/lib/mdtut.server.ts
@@ -57,7 +57,7 @@ async function getMarkdownTutPage(filename: string): Promise<MarkdownTutPage> {
   if (!file) {
     throw new Error('File not found: "' + filename + '"');
   }
-  let page = (await processMarkdown(processor, file)) as MarkdownTutPage;
+  let page = (await processMarkdown(processor as TProcessor, file)) as MarkdownTutPage;
   cache.set(filename, page);
   return page;
 }
@@ -68,6 +68,7 @@ type TProcessor = Processor<
   Hast.Node<Unist.Data>,
   Hast.Node<Unist.Data>,
   Hast.Node<Unist.Data>,
+  Unist.Parent,
   string
 >;
 
@@ -168,7 +169,7 @@ function stringify(
         let html = processor
           .stringify({ type: "root", children: childNodes } as Mdast.Root)
           .trim();
-        let subject = processor.stringify(subjectNode).trim();
+        let subject = processor.stringify(subjectNode as Mdast.Root).trim();
         slides.push({ type: "slide", subject, html });
       }
       page.push({ type: "sequence", slides });
--- a/app/lib/mdtut.server.ts
+++ b/app/lib/mdtut.server.ts
@@ -1,6 +1,5 @@
 import { loadPlugins, type UnistNode } from "~/lib/md.server";
 import { LRUCache } from "lru-cache";
-import type { Processor } from "unified";
 import type * as Hast from "hast";
 import type * as Mdast from "mdast";
 import type * as Unist from "unist";
@@ -25,7 +24,8 @@ const cache = new LRUCache<string, MarkdownTutPage>({
   },
 });
 
-let processorPromise: ReturnType<typeof getProcessor>;
+type TProcessorPromise = ReturnType<typeof getProcessor>;
+let processorPromise: TProcessorPromise;
 
 async function getProcessor() {
   let [
@@ -57,27 +57,21 @@ async function getMarkdownTutPage(filename: string): Promise<MarkdownTutPage> {
   if (!file) {
     throw new Error('File not found: "' + filename + '"');
   }
-  let page = (await processMarkdown(processor as TProcessor, file)) as MarkdownTutPage;
+  let page = (await processMarkdown(processor, file)) as MarkdownTutPage;
   cache.set(filename, page);
   return page;
 }
 
 ////////////////////////////////////////////////////////////////////////////////
 
-type TProcessor = Processor<
-  Hast.Node<Unist.Data>,
-  Hast.Node<Unist.Data>,
-  Hast.Node<Unist.Data>,
-  Unist.Parent,
-  string
->;
+type TProcessor = Awaited<TProcessorPromise>;
 
 async function processMarkdown(
   processor: TProcessor,
   content: string | Buffer,
 ): Promise<MarkdownTutPage> {
   let state: State = STATE_NORMAL;
-  let tree = (await processor.run(processor.parse(content))) as UnistNode.Root;
+  let tree = await processor.run(processor.parse(content) as UnistNode.Root);
   let current: ProseNode | SlideNode = {
     type: "prose",
     children: [],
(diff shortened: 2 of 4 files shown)


brookslybrand (approved):
Thanks so much @takagimeow!

I poked around in the docs and blog (places where we parse code in markdown) and everything looked the same 👍

Thanks for the thorough comments. I was curious with the issues in `mdtut.server.ts`, did you attempt to update `TProcessor` to avoid the type casting? Or was that just too complicated. Personally I don't care too much, but it does seem like we're adding more bandaids to the types, making the types less useful for us in the future. Not a show stopper, just something I was curious about

takagimeow:
> Thanks so much @takagimeow!
> 
> I poked around in the docs and blog (places where we parse code in markdown) and everything looked the same 👍
> 
> Thanks for the thorough comments. I was curious with the issues in `mdtut.server.ts`, did you attempt to update `TProcessor` to avoid the type casting? Or was that just too complicated. Personally I don't care too much, but it does seem like we're adding more bandaids to the types, making the types less useful for us in the future. Not a show stopper, just something I was curious about


Thank you for checking on it! 🙏

The reason for updating `TProcessor` was due to errors encountered when passing `Hast.Node<Unist.Data>` as the fourth type argument, `CompileTree`, to `TProcessor`, as was done previously.

```shell
app/lib/mdtut.server.ts:60:37 - error TS2352: Conversion of type 'Processor<Root, UnistNode.Root, UnistNode.Root, Root, string>' to type 'TProcessor' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Property 'children' is missing in type 'Node<Data>' but required in type 'Root'.

60   let page = (await processMarkdown(processor as TProcessor, file)) as MarkdownTutPage;
                                       ~~~~~~~~~~~~~~~~~~~~~~~

  node_modules/remark-gfm/node_modules/@types/mdast/index.d.ts:506:5
    506     children: RootContent[];
            ~~~~~~~~
    'children' is declared here.


Found 1 error in app/lib/mdtut.server.ts:60
```

The `processor` is defined as `Processor<Root, UnistNode.Root, UnistNode.Root, Root, string>`. Here, the fourth type argument, `Root`, refers to `mdast`'s `Root`.

To resolve this error and ensure the constraints of abstract types, `UnistNode`, which is a supertype of `mdast`'s `Root`, was passed instead of `Hast.Node<Unist.Data>`.

As you pointed out, the more complex the types become, the higher the maintenance cost, potentially leading to less usefulness in the future.

Given how `TProcessor` is used, I think it would be better defined as an alias for `Awaited<ReturnType<typeof getProcessor>>` rather than defining it using `Processor`.

brookslybrand:
> Given how `TProcessor` is used, I think it would be better defined as an alias for `Awaited<ReturnType<typeof getProcessor>>` rather than defining it using `Processor`.

That makes sense to me! I'm all for updating the types however you think makes the most sense. I'd prefer to get rid of type casting here as much as possible, since it just seem to be hiding implementation details that you have to relearn each time you look at the code by parsing through the flow manually.

takagimeow:
Thank you so much @brookslybrand!

I updated `TProcessor` as an alias for the awaited return type of `getProcessor()`.

This eliminates the need for type casting when passing `processor` to `processMarkdown()`.

The implementation now feels more transparent and straightforward compared to before 😊.

see: https://github.com/remix-run/remix-website/pull/174/commits/2a9167dc9f57d3e46b7911bca688a9c4abef1104

Additionally, I explicitly installed `@types/mdast` as a devDependency. 

see: https://github.com/remix-run/remix-website/pull/174/commits/8d1fb12c9acad4a2c6c732b2fc5d16dceaefa397

This is because the version of `@types/mdast` that `remark-parse` and `remark-html` depend on differs from the version in `node_modules/@types/mdast`.

- [remark-html@16.0.1 package.json](https://github.com/remarkjs/remark-html/blob/b63ddf4d1828ea28ce3f232c19236176dfb5f549/package.json#L40)
- [remark-parse@11.1.0 package.json](https://github.com/remarkjs/remark/blob/618a9ad1d44aa106bd2a8c61ebf8589cfe97fa16/packages/remark-parse/package.json#L43)

The `node_modules/@types/mdast` is likely installed by `remark-parse` depended upon by `@mdx-js/mdx`, which `remix-run/dev` relies on.

- [remix-run/dev@2.5.1 package.json](https://github.com/remix-run/remix/blob/043d14df13d3830cf7a83bc06817de7cac92c2f3/packages/remix-dev/package.json#L29)
- [mdx-js/mdx@2.3.0 package.json](https://github.com/mdx-js/mdx/blob/b117f749d701d47d23efa0903579c49df685cf8d/packages/mdx/package.json#L59)
- [remark-parse@10.0.0 package.json](https://github.com/remarkjs/remark/blob/9a09d0982347da0cba003b14e67c1d883cc7983f/packages/remark-parse/package.json#L44)

The version of `@mdx-js/mdx` that `remix-run/dev` depends on is `2.3.0`, which in turn depends on version `10.0.0` of `remark-parse`. This version of `remark-parse` depends on version `3.0.0` of `@types/mdast`. However, the updated `remark-parse` and `remark-html` depend on version `4.0.0` of `@types/mdast`, leading to type errors.

Since the latest version of `@remix-run/dev` still depends on version `2.3.0` of `@mdx-js/mdx`, the only way to resolve this type error at this time is to explicitly install `@types/mdast` as a devDependency.

- [remix-run/dev@2.6.0 package.json](https://github.com/remix-run/remix/blob/2423f6511585d82ed537b46e3d1a1526fc7aeb4c/packages/remix-dev/package.json#L29)

brookslybrand (approved):
Awesome! Yes, this looks a lot cleaner, thank you @takagimeow 🎉

Issue #109, opened 2023-09-28
Add types/validation for matches


brookslybrand:
`useMatches` and other cases of matches don't supply any types. With the v2 upgrade, this ended up causing TS to start failing. Added a bunch of `@ts-expect-error` statements with the intention of coming back and fixing them

If using `zod` makes this easy, that's great since we already have it in this project. However, avoid using it in any data that ships to the client (so server-side only)

Kazuhiro-Mimaki:
Could I do this?

takagimeow:
@brookslybrand 
Since the PR related to this issue has been merged, I believe this issue should be closed!

brookslybrand:
Yep! Thanks for pointing that out

Fixed in #175

PR #175, opened 2024-02-04
Add types/validation for matches

Kazuhiro-Mimaki:
In this PR, I added zod validation for matches and remove `ts-expect-error` annotation.

## Related
- https://github.com/remix-run/remix-website/issues/109

### DIFF
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -142,8 +142,12 @@ function Document({
 export default function App() {
   let matches = useMatches();
   let { noIndex } = useLoaderData<typeof loader>();
-  // @ts-expect-error -- useMatches types changed to `unknown`, need to validate
-  let forceDark = matches.some((match) => match.handle?.forceDark);
+  let forceDark = matches.some(({ handle }) => {
+    if (handle && typeof handle === "object" && "forceDark" in handle) {
+      return handle.forceDark;
+    }
+    return false;
+  });
 
   if (process.env.NODE_ENV !== "development") {
     // eslint-disable-next-line react-hooks/rules-of-hooks
--- a/app/routes/conf.2023.tsx
+++ b/app/routes/conf.2023.tsx
@@ -27,6 +27,7 @@ import {
   SubscribeSubmit,
 } from "~/ui/subscribe";
 import { CACHE_CONTROL } from "~/lib/http.server";
+import invariant from "tiny-invariant";
 
 export const handle = { forceDark: true };
 
@@ -223,28 +224,27 @@ function isExternalUrl(value: string, currentHost: string) {
   }
 }
 
+function hasHost(data: unknown): data is { host: string } {
+  return !!data && typeof data === "object" && "host" in data;
+}
+
 const HeaderLink = React.forwardRef<HTMLAnchorElement, HeaderLinkProps>(
   ({ to, children, className, prefetch = "none", ...props }, ref) => {
-    let rootMatch = useMatches().find((match) => match.id === "root")!;
-    if (
-      rootMatch.data &&
-      typeof rootMatch.data === "object" &&
-      "host" in rootMatch.data &&
-      typeof rootMatch.data.host === "string"
-    ) {
-      if (isExternalUrl(to, rootMatch.data.host)) {
-        return (
-          <a
-            ref={ref}
-            className={cx("text-base font-semibold", className)}
-            href={to}
-            children={children}
-            target="_blank"
-            rel="noopener noreferrer"
-            {...props}
-          />
-        );
-      }
+    let rootMatchData = useMatches().find((match) => match.id === "root")?.data;
+    invariant(hasHost(rootMatchData), "No host found in root match data");
+    let external = isExternalUrl(to, rootMatchData.host);
+    if (external) {
+      return (
+        <a
+          ref={ref}
+          className={cx("text-base font-semibold", className)}
+          href={to}
+          children={children}
+          target="_blank"
+          rel="noopener noreferrer"
+          {...props}
+        />
+      );
     }
 
     return (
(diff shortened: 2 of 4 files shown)


app/root.tsx
```diff
@@ -137,11 +138,19 @@ function Document({
   );
 }
 
+const matchHandleSchema = z
+  .object({
+    forceDark: z.boolean(),
+  })
+  .optional();
+
```
brookslybrand, line 146: Sorry, I should have been more clear in the issue. I'd prefer not to ship `zod` to the client, since it's [kind of large](https://bundlephobia.com/package/zod@3.22.4).

All good to use it on the server as we already are. This is kind of the challenge with fullstack frameworks that try to blend the client and the server, it's easy to leak stuff like this. Would probably be good to put all zod validation stuff inside of a directory called `validate.server/` or something

Here's a simple, type-safe validation that doesn't require sending `zod` to the client. I get it's a bit verbose, so you're welcome to extract it to a function or figure out a better approach

```js
let forceDark = matches.some(({ handle }) => {
  if (handle && typeof handle === "object" && "forceDark" in handle) {
    return handle.forceDark;
  }
  return undefined;
});
```

app/routes/conf.2023.tsx
```diff
@@ -223,11 +224,15 @@ function isExternalUrl(value: string, currentHost: string) {
   }
 }
 
+const rootMatchDataSchema = z.object({
+  host: z.string(),
+});
```
brookslybrand, line 229: Same comment here, we want to avoid shipping `zod` to the client

app/routes/docs.$lang.$ref.tsx
```diff
@@ -755,10 +756,47 @@ function InnerContainer({ children }: { children: React.ReactNode }) {
...
+      headings: z.array(
+        z
+          .object({
+            html: z.string().nullable(),
+            slug: z.string().optional(),
+          })
+          .transform((v) => ({ ...v, slug: v.slug })),
+      ),
+    }),
+  });
```
brookslybrand, line 792: Thanks for working on this, this is pretty robust checking. Again, I want to avoid sending `zod` to the client, but I also kind of want to avoid doing a bunch of client side logic to parse and validate every field on the doc.

I think we can just do some inferring and create a little type guard here:

```
function hasDoc(data: unknown): data is { doc: Doc } {
  return !!data && typeof data === "object" && "doc" in data;
}

function useDoc(): Doc {
  const data = useMatches().at(-1)?.data;
  invariant(hasDoc(data), "No doc data found in loader data");
  return data.doc;
}
```

I'm open to alternative approaches, basically I just don't think there's a big reason to parse the entire data set each time when we can pretty safely assume if there is a `doc` field it's going to be the appropriate type

app/root.tsx
kiliman: If you're using Vite, then you can use the `serverOnly$` export from `vite-env-only` to ensure that zod is not included in the client bundle.

```ts
// routes/test.$id.tsx

import { json, type LoaderFunctionArgs } from '@remix-run/node'
import { useLoaderData } from '@remix-run/react'
import { serverOnly$ } from 'vite-env-only'
import { z } from 'zod'

// this ensures that zod is only included in the server bundle
const schema = serverOnly$(() => z.object({ id: z.coerce.number() }))

export async function loader({ params }: LoaderFunctionArgs) {
  const { id } = schema!().parse(params)
  return json({ id })
}

export default function Component() {
  const loaderData = useLoaderData<typeof loader>()

  return (
    <div>
      <pre>{JSON.stringify(loaderData, null, 2)}</pre>
    </div>
  )
}
```

Kazuhiro-Mimaki:
Thank you review.
I'm sorry but I misunderstand.
Your pretty clear explanation and useful examples help me to understand what to do.
I remove `zod` from client, could you review my code again?

brookslybrand:
> Thank you review. I'm sorry but I misunderstand. Your pretty clear explanation and useful examples help me to understand what to do. I remove `zod` from client, could you review my code again?

Wonderful! And no need to apologize, my original issue wasn't super clear. Looking again now 👀

app/routes/docs.$lang.$ref.$.tsx
```diff
@@ -59,6 +59,28 @@ export const headers: HeadersFunction = ({ loaderHeaders }) => {
...
+  );
+}
+
+function hasIsProductionHost(
+  data: unknown,
+): data is { isProductionHost: boolean } {
+  return !!data && typeof data === "object" && "isProductionHost" in data;
+}
+
 export const meta: MetaFunction<
```
brookslybrand, line 84: Alright, I just realized something. We're actually declaring the types of `matches` data and passing it into `MetaFunction`, however those types aren't preserved by `getMatchesData`. If we just update this slightly we get all of the types we want. Since they're coming from the actual types of the `loaders`, this is about as much type safety as we need

```ts
const LAYOUT_LOADER_KEY = "routes/docs.$lang.$ref";

type Loader = typeof loader;
type MatchLoaders = {
  [LAYOUT_LOADER_KEY]: typeof docsLayoutLoader;
  root: typeof rootLoader;
};
export const meta: MetaFunction<Loader, MatchLoaders> = (args) => {
  let { data } = args;

  let matchesData = getMatchesData<Loader, MatchLoaders>(args);
```

`matchesData` will be typed properly at this point.

I think this is the only suggestion I have at this point, otherwise I'm good to merge it in!

Kazuhiro-Mimaki: Thank you for your suggestion.
Passing types to `getMatchesData` makes the code simple, it is very nice!

brookslybrand (approved):
Awesome, thanks @Kazuhiro-Mimaki for responding to all the feedback! Merging in now 😎

Issue #166, opened 2024-02-02
Bug: npm let us use undeclared (in package.json) dependencies


fernandocanizo:
**Is there a serious showstopper reason to not be using `pnpm` yet?**
(jump to the end to see about the bug from the title of this issue)

- For all these excerpts, assume `$ rm -rf node_modules/` has been run prior to the timed command.
- First run was without lock file and second run was always with lock file.
- Verification made while standing on HEAD, which currently is commit `f54dbcd`

## npm

**First run: ~55 seconds**
```
$ time npm ci

added 1261 packages, and audited 1262 packages in 54s

10 vulnerabilities (9 moderate, 1 high)

To address issues that do not require attention, run:
  npm audit fix

To address all issues (including breaking changes), run:
  npm audit fix --force

Run `npm audit` for details.

real    0m54,574s
user    0m18,137s
sys     0m15,546s
```

**Second run: ~20 seconds**
```
$ time npm ci
[redacted...]
real    0m19,637s
user    0m11,526s
sys     0m8,368s
```

## pnpm

**First run: ~40 seconds**
```
$ time pnpm i
Packages: +1168
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Progress: resolved 1232, reused 1167, downloaded 0, added 1168, done

[redacted...]

Done in 39.5s

real    0m39,637s
user    0m16,913s
sys     0m14,826s
```

**Second run: ~3 seconds**
```
$ time pnpm i
[redacted...]
Done in 2.7s

real    0m2,802s
user    0m8,708s
sys     0m2,670s
```

## yarn

**First run: ~113 seconds**
```
$ time yarn install
yarn install v1.22.21
info No lockfile found.
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
[4/5] Linking dependencies...
warning "@docsearch/react > @algolia/autocomplete-preset-algolia@1.9.3" has unmet peer dependency "@algolia/client-search@>= 4.9.1 < 6".
warning "@docsearch/react > @algolia/autocomplete-core > @algolia/autocomplete-plugin-algolia-insights@1.9.3" has unmet peer dependency "search-insights@>= 1 < 3".
[...]

PR #178, opened 2024-02-05
Add `react-router-dom` package

fernandocanizo:
Due to the way `npm` solves peer dependencies making a flat `node_modules`, we were able to import `react-router-dom` in the project even when we didn't have it as dependency.

This adds the dependency to avoid possible future bugs for realying on an undeclared dependency.

Also, bumps `react-router-dom` version from 6.21.3 to 6.22.0

While this is not a switch to `pnpm`, it fixes the bug mentioned in #166


fernandocanizo:
Superseded by #179

PR #179, opened 2024-02-05
Import `matchPath` from declared dependency

fernandocanizo:
`matchPath` was being imported from an undeclared dependency, but could be used due to `npm` flattening peer-dependencies. This way we ensure we're using the expected version.


brookslybrand (approved):
Thanks @fernandocanizo!

PR #181, opened 2024-02-07
fix: editUrl

yikayiyo:
image


brookslybrand (approved):
Oh shoot, thanks for catching this!

We broke this in #165. Less hard coding is better anyway 🤘

PR #184, opened 2024-02-07
feat: add /blog/rss.xml route

### LINKED ISSUE #182: Add RSS feed?
I'm surprised that the Remix does not have an RSS feed.
(but you are not alone see https://twitter.com/sebastienlorber/status/1754927448641135099)

Having an RSS feed would be useful for newsletter curators like me to notice new posts as soon as they are published, and feature them.

Alternatives are less "real-time" than having a proper RSS feed on your site:
- the Remix newsletter is not sent immediately after a new post
- the [Remix.guide](https://remix.guide/) site is apparently updated in batches
- using RSS reader "web feed" features is usually a paid option to track websites not having proper RSS feeds.

takagimeow:
closes #182 

As mentioned by @slorber and @brookslybrand , I also think that Remix should have an RSS feed for blog posts. Therefore, I updated it so that an appropriate RSS feed can be visible by accessing /blog/rss.xml.

### DIFF
--- a/app/routes/_extras.blog.rss[.xml].tsx
+++ b/app/routes/_extras.blog.rss[.xml].tsx
@@ -1,0 +1,46 @@
+import type { LoaderFunction } from "@remix-run/node";
+import { getBlogPostListings } from "~/lib/blog.server";
+
+const getCData = (text: string) => {
+  return `<![CDATA[${text}]]>`;
+};
+
+export const loader: LoaderFunction = async () => {
+  const blogUrl = `https://remix.run/blog`;
+  const posts = await getBlogPostListings();
+
+  const rss = `
+    <rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom">
+      <channel>
+        <title>Remix Blog</title>
+        <link>${blogUrl}</link>
+        <atom:link href="${blogUrl}/rss.xml" rel="self" type="application/rss+xml" />
+        <description>Thoughts about building excellent user experiences with Remix.</description>
+        <language>en-us</language>
+        <generator>https://github.com/remix-run/remix</generator>
+        <copyright>© Shopify, Inc.</copyright>
+        ${posts
+          .map(post =>
+            `
+            <item>
+              <title>${getCData(post.title)}</title>
+              <description>RFC-822 date: ${getCData(post.summary)}</description>
+              <pubDate>${new Date(post.dateDisplay).toUTCString()}</pubDate>
+              <link>${blogUrl}/${post.slug}</link>
+              <guid>${blogUrl}/${post.slug}</guid>
+            </item>
+          `.trim(),
+          )
+          .join('\n')}
+      </channel>
+    </rss>
+  `.trim();
+
+
+  return new Response(rss, {
+    headers: {
+      "Content-Type": "application/xml",
+      "Cache-Control": "public, max-age=0",
+    },
+  });
+};
--- a/app/routes/_extras.blog.rss[.xml].tsx
+++ b/app/routes/_extras.blog.rss[.xml].tsx
@@ -1,6 +1,7 @@
 import type { LoaderFunction } from "@remix-run/node";
 import { getBlogPostListings } from "~/lib/blog.server";
 import { Feed } from "feed";
+import { CACHE_CONTROL } from "~/lib/http.server";
 
 export const loader: LoaderFunction = async () => {
   const blogUrl = `https://remix.run/blog`;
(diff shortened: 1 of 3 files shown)


app/routes/_extras.blog.rss[.xml].tsx
```diff
@@ -0,0 +1,46 @@
...
+        <description>Thoughts about building excellent user experiences with Remix.</description>
+        <language>en-us</language>
+        <generator>https://github.com/remix-run/remix</generator>
+        <copyright>© Shopify, Inc.</copyright>
+        ${posts
+          .map(post =>
+            `
+            <item>
+              <title>${getCData(post.title)}</title>
+              <description>RFC-822 date: ${getCData(post.summary)}</description>
```
Comment (brookslybrand) [inline app/routes/_extras.blog.rss[.xml].tsx:27]: I'm assuming this just got left behind from wherever you took this code?

```suggestion
              <description>${getCData(post.summary)}</description>
```

app/routes/_extras.blog.rss[.xml].tsx
```diff
@@ -0,0 +1,46 @@
...
+          .join('\n')}
+      </channel>
+    </rss>
+  `.trim();
+
+
+  return new Response(rss, {
+    headers: {
+      "Content-Type": "application/xml",
+      "Cache-Control": "public, max-age=0",
```
Comment (brookslybrand) [inline app/routes/_extras.blog.rss[.xml].tsx:43]: Any reason to not add a max-age value here? I've never setup an RSS feed, so there very well may be a reason. Otherwise I would just do the following

```suggestion
      "Cache-Control": CACHE_CONTROL.DEFAULT
```

You'll need to import CACHE_CONTROL of course

brookslybrand (changes requested):
@takagimeow you're on fire, thank you 🙏 

Just a couple of comments

Also, if you borrowed this code from somewhere, would you put an attribution/link in a comment at the top?

app/routes/_extras.blog.rss[.xml].tsx
Comment (takagimeow) [reply app/routes/_extras.blog.rss[.xml].tsx]: Thank you very much for pointing that out @brookslybrand!

I used the [W3C Feed Validation Service](https://validator.w3.org/feed/check.cgi) to verify if the RSS feed generated was correct and encountered an error as follows.

Initially, I set `<pubDate>` using `toISOString()`, which led to this error. 

スクリーンショット 2024-02-08 10 12 33

In the process of resolving this error, I experimented with various date and time formats to see which would not cause an error. Somehow, and I don't know why, I got confused between `<description>` and `<pubDate>` and kept trying through trial and error.

So, this part was simply something I forgot to remove from when I fixed that error.! I'm truly sorry for the inconvenience caused. 🙇

see: https://github.com/remix-run/remix-website/pull/184/commits/0d217880ccd8ae66c62bb9b34e194e026f40328e

app/routes/_extras.blog.rss[.xml].tsx
Comment (takagimeow) [reply app/routes/_extras.blog.rss[.xml].tsx]: I'm truly sorry! I hadn't realized the existence of `CACHE_CONTROL.default` 🙇!
I fixed it now 👍

https://github.com/remix-run/remix-website/pull/184/commits/41b72c6290cc709f0f453467b4efb4d588343ef2

takagimeow:
@brookslybrand 

Human input errors like [this](https://github.com/remix-run/remix-website/pull/184#discussion_r1482103756) need to be prevented, and as I referenced various codes when creating the body of the RSS, it became necessary to clarify the licensing issues of the code. Therefore, I updated the code to generate RSS using [jpmonette/feed](https://github.com/jpmonette/feed)!

I kindly request you to review it once again 🙏.

app/routes/_extras.blog.rss[.xml].tsx
Comment (brookslybrand) [reply app/routes/_extras.blog.rss[.xml].tsx]: Haha no worries, I figured it was something like that. Thanks so much for validating the output

app/routes/_extras.blog.rss[.xml].tsx
Comment (brookslybrand) [reply app/routes/_extras.blog.rss[.xml].tsx]: Haha no need to apologize, you had no reason to know about this. Thanks for the update 🙂

I'll do a little more checking and then merge this at some point today 🎉

brookslybrand (approved):

slorber:
The feed is now there but it can't be automatically discovered, you need a meta for that:

```html
<link rel="alternate" type="application/rss+xml" href="/..." />
```

brookslybrand:
Ah, thanks @slorber for pointing that out. If it's not obvious I don't use RSS 😅 

Fixed in #187

takagimeow:
@brookslybrand 
Thank you so much for fixing that issue 🙏

PR #185, opened 2024-02-07
Correct remix-auth-microsoft initCommand

arnabsen:


brookslybrand (approved):
Whoops, thanks for catching this!

PR #191, opened 2024-02-09
fix: resolve symlinked files

### LINKED ISSUE #190: Bug: Resources README doesn't always render correctly
See [Remix Fastify](https://remix.run/resources/remix-fastify) for an example

When I wrote the README fetcher for the Resource pages, I took a [naive approach](https://github.com/remix-run/remix-website/blob/main/app/lib/resources.server.ts#L107-L114)

Basically I'm assuming that the README url is always `main/readme.md`. This is not always true (see the Remix Fastify example).

The good thing is `octokit` will give you the appropriate URL with the following

```js
context.octokit.rest.repos.getReadme({ owner, repo })
```

All that's required is a little refactoring of code to fetch the correct content and the rendering should take care of itself

mcansh:
mcansh/remix-fastify root README is a symlink to the README in packages/remix-fastify and using the raw.github... doesnt follow the redirect

closes #190

feel free to close if any of the other commenters on the issue also do the work so they can get the contribution :)

Signed-off-by: Logan McAnsh <logan@mcan.sh>

### DIFF
--- a/app/lib/gh-docs/docs.ts
+++ b/app/lib/gh-docs/docs.ts
@@ -6,6 +6,9 @@ import { getRepoTarballStream } from "./repo-tarball";
 import { createTarFileProcessor } from "./tarball.server";
 import { load as $ } from "cheerio";
 import { env } from "~/env.server";
+import type { Octokit } from "octokit";
+
+type CacheContext = { octokit: Octokit };
 
 interface MenuDocAttributes {
   title: string;
@@ -29,7 +32,7 @@ export interface Doc extends Omit<MenuDoc, "hasContent"> {
 
 declare global {
   var menuCache: LRUCache<string, MenuDoc[]>;
-  var docCache: LRUCache<string, Doc>;
+  var docCache: LRUCache<string, Doc, CacheContext>;
 }
 
 let NO_CACHE = env.NO_CACHE;
@@ -81,8 +84,7 @@ function parseAttrs(
  * let's have simpler and faster deployments with just one origin server, but
  * still distribute the documents across the CDN.
  */
-global.docCache ??= new LRUCache<string, Doc>({
-  // let docCache = new LRUCache<string, Doc | undefined>({
+global.docCache ??= new LRUCache<string, Doc, CacheContext>({
   max: 300,
   ttl: NO_CACHE ? 1 : 1000 * 60 * 5, // 5 minutes
   allowStale: !NO_CACHE,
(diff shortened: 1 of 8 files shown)


app/lib/gh-docs/repo-content.ts
```diff
@@ -17,13 +18,25 @@ export async function getRepoContent(
...
   if (ref === "local") return getLocalContent(filepath);
   let [owner, repo] = repoPair.split("/");
-  let pathname = `/${owner}/${repo}/${ref}/${filepath}`;
-  let response = await fetch(
-    new URL(pathname, "https://raw.githubusercontent.com/").href,
-    { headers: { "User-Agent": `docs:${owner}/${repo}` } },
-  );
-  if (!response.ok) return null;
-  return response.text();
+  let contents = await octokit.rest.repos.getContent({
```
brookslybrand, line 21: This is probably fine to pull in `octokit` directly. However, I noticed that in [some other files](https://github.com/remix-run/remix-website/blob/main/app/lib/gh-docs/tags.ts#L16_ we pass `octokit` in as an argument. Do you know if there was any specific reason for this pattern initially?

mcansh: i don't know if there was any reason for passing it, looks like chance made that decision sometime last year before he left. i can match the pattern if preferred.

app/lib/gh-docs/docs.ts
```diff
@@ -81,19 +82,22 @@ function parseAttrs(
...
   ttl: NO_CACHE ? 1 : 1000 * 60 * 5, // 5 minutes
   allowStale: !NO_CACHE,
   noDeleteOnFetchRejection: true,
   fetchMethod: fetchDoc,
 });
 
-async function fetchDoc(key: string): Promise<Doc> {
+async function fetchDoc(
+  key: string,
+  _staleValue: Doc | undefined,
```
brookslybrand, line 95: I'm not sure I understand why `_staleValue` is in here, doesn't seem like we use it. Also, when we call `docCache.fetch` we're not passing it in, am I missing something?

app/lib/resources.server.ts
```diff
@@ -96,26 +95,35 @@ declare global {
...
+async function fetchReadme(
+  key: string,
+  _staleValue: string | undefined,
+  { context }: LRUCache.FetchOptionsWithContext<string, string, CacheContext>,
+): Promise<string> {
+  let [owner, repo] = key.split("/");
+  let contents = await context.octokit.rest.repos.getReadme({
+    owner,
+    repo,
+    mediaType: { format: "base64" },
```
brookslybrand, line 115: I'm just now realizing that we can just make it `format: "raw"` 🤦. I feel like that makes sense, what do you think? Pretty sure raw will also be smaller in size and less work than getting a base64 encoding

app/lib/gh-docs/docs.ts
mcansh: it’s a required argument when using the LRU.FetchOptionsWithContext as we’re passing around an octokit instance, but yes, we’re not using it, it’s added by LRUCache. followed convention from other cache.fetch functions

brookslybrand: Great, figured I was missing something, thanks for the explanation!

brookslybrand (approved):
Great improvements on top of a fix. Thanks @mcansh!

PR #192, opened 2024-02-12
Allow custom blog post author order

markdalgleish:
Blog post authors are currently listed in alphabetical order. This change ensures that blog post authors are listed in the order they were defined in the post's frontmatter.

### DIFF
--- a/app/lib/blog.server.ts
+++ b/app/lib/blog.server.ts
@@ -15,6 +15,7 @@ const postContentsBySlug = Object.fromEntries(
 );
 
 const AUTHORS: BlogAuthor[] = yaml.parse(authorsYamlFileContents);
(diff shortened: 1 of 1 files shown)


app/lib/blog.server.ts
```diff
@@ -84,14 +85,8 @@ function getAuthor(name: string): BlogAuthor | undefined {
   return AUTHORS.find((a) => a.name === name);
 }
 
-function getValidAuthorNames(authors: string[]) {
-  let validatedAuthors: string[] = [];
-  for (let author of AUTHORS) {
-    if (authors.includes(author.name)) {
-      validatedAuthors.push(author.name);
```
markdalgleish, line 91: The alphabetical order was a side effect of iterating over the complete author list rather than filtering the input author list.

PR #203, opened 2024-02-21
Render docs search bar on all branches

### LINKED ISSUE #200: Render the docs search bar on all branches
## tl;dr

Render the docs search bar on all docs (`main`, `dev`, previous versions, etc.)

## Context

Recently I shared [a tweet](https://twitter.com/BrooksLybrand/status/1758601575000375496) hyping up some changes coming to the next version of Remix.

This highlighted the fact that switching between branches causes the search bar to disappear.

[I summarized the reason and the remediation options](https://twitter.com/BrooksLybrand/status/1759696092986290303) as follows

> It's happening because we only render the search box for `main` branch. We only render it for `main` because that's the only docs we index.

> So either we need to:
> a) index other branches
> b) just show the search box on every branch, even though it's only searching main

For the sake of consistency between branches, let's go with **b** for now. If we find that this causes confusion or frustration for users we can further enhance it with some sort of message or by implementing **a**

mingjunlu:
## Changes

- Renders docs search bar on all branches by removing [these 2 lines](https://github.com/remix-run/remix-website/blob/main/app/routes/docs.%24lang.%24ref.tsx#L346-L347)
- Closes #200 

## Demonstration

| Before | After |
| ------ | ----- |
| before | after |

### DIFF
--- a/app/routes/docs.$lang.$ref.tsx
+++ b/app/routes/docs.$lang.$ref.tsx
@@ -343,8 +343,6 @@ function VersionLink({
 }
 
 function DocSearchSection({ className }: { className?: string }) {
-  let { isLatest } = useLoaderData<typeof loader>();
-  if (!isLatest) return null;
   return (
     <div className={cx("relative lg:sticky lg:top-0 lg:z-10", className)}>
       <div className="absolute -top-24 hidden h-24 w-full bg-white dark:bg-gray-900 lg:block" />


brookslybrand (approved):
Looks good to me! Now we just have to wait until someone complains about search on dev/old docs pointing to the latest 🙃

PR #212, opened 2024-02-27
fix: Fixed broken urls in READMEs

### LINKED ISSUE #211: Fix Resource README's with relative images
If a Resource's README uses a relative link in the README source code, [our rendering](https://remix.run/resources/remix-pwa) isn't able to load the image

image

- [Example repo with rendered README](https://github.com/remix-pwa/remix-pwa)
- [source code](https://raw.githubusercontent.com/remix-pwa/remix-pwa/main/README.md)
- [Code to fetch the README]: https://github.com/remix-run/remix-website/blob/main/app/lib/resources.server.ts#L107-L126

I'm not sure if there's a good way with `octokit` to fix this, but if there is that would be the preference, rather than parsing over all the entire README and trying to find relative images

AlemTuzlak:
Fixes #211

This will fix the broken URL's in resources.

I added "ES2022" as a lib which I'm not sure if this is okay, if not this can also be done with good old replace.


brookslybrand:
Hmm, yeah I'm a little worried about this approach. If there are every code samples that use a relative path this is going to break those examples 😕

I don't know if you looked into `octokit` at all to see if it has APIs for resolving the relative links, I would maybe do a little research there before trying to implement something ourselves.

If we do take our own find-and-replace approach, I think this is definitely something we should provide some tests for, since I'm sure we'll run into bugs as we go. Don't want to play bug wack-a-mole.

tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "include": ["env.d.ts", "**/*.ts", "**/*.tsx"],
   "compilerOptions": {
-    "lib": ["DOM", "DOM.Iterable", "ES2019"],
+    "lib": ["DOM", "DOM.Iterable", "ES2019", "ES2022"],
```
brookslybrand, line 4: I'm fine with updating the `"lib"`, but we don't need to have `"ES2019"` and `"ES2022"`

AlemTuzlak:
After checking the docs here:
https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#get-a-repository-readme

I tried using json and raw, after trying both they return relative links instead of the absolute ones. I think that ocktokit does not support these kind of replacements at the moment. 

I've enhanced the function to use a regex instead that finds the img attribute and runs the replacement, also added a bunch of tests for all cases I could think of. Let me know what you think @brookslybrand

app/lib/resources.server.ts
```diff
@@ -80,7 +110,11 @@ export async function getResource(
     throw new Error(`Could not find GitHub data for ${resource.repoUrl}`);
   }
 
-  return { ...resource, ...gitHubData, readmeHtml };
+  return {
+    ...resource,
+    ...gitHubData,
+    readmeHtml: replaceRelativeLinks(readmeHtml, resource.repoUrl),
```
brookslybrand, line 116: Let's move the `replaceRelativeLinks` inside `fetchReadme`, since the output of `fetchReadme` is server-side cached in the LRU cache (`resourceReadmeCache`). Otherwise we're going to do this replacement every time the resources is requested, instead of just the first time (every 5 minutes).

This should do it

```
let { html } = await processMarkdown(md);
return replaceRelativeLinks(html, `https://github.com/${key}`);
```

Feel free to pull `https://github.com/` out as a constant and reference that, since we use it like 3 places. No big deal if you don't want to do that step

app/lib/tests/resources.server.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    const input = `<img src="./relative1"><img src="./relative2">`;
+    const repoUrl = "https://my-repo";
+    const expected = `<img src="${repoUrl}/raw/main/relative1"><img src="${repoUrl}/raw/main/relative2">`;
+    expect(replaceRelativeLinks(input, repoUrl)).toEqual(expected);
+  });
+
+  it("should replace relative links with absolute links for images with other attributes", () => {
+    const input = `<img src="./relative" alt="alt text">`;
+    const repoUrl = "https://my-repo";
+    const expected = `<img src="${repoUrl}/raw/main/relative" alt="alt text">`;
```
brookslybrand, line 28: Unfortunately this breaks if `alt` (or any attribute) comes before `img`

```js
    const input = `<img alt="alt text" src="./relative">`;
    const repoUrl = "https://my-repo";
    const expected = `<img alt="alt text" src="${repoUrl}/raw/main/relative">`;
```

brookslybrand (changes requested):
Great stuff Alem! Just a small tweak and a test case update

Also, thanks so much for all the thorough tests! I do think you can collapse a few of these tests together. For example, the test with multiple images with extra attributes mixed with images with non-relative links covers about 4-5 tests. If a test is already testing other things, I think it's safe to remove at least some of the others

app/lib/resources.server.ts
AlemTuzlak: good point here!

app/lib/tests/resources.server.test.ts
AlemTuzlak: good catch, added a failing test, then changed it to work!

app/lib/resources.server.ts
```diff
@@ -8,6 +8,8 @@ import type { Octokit } from "octokit";
 
 export type CacheContext = { octokit: Octokit };
 
+const GITHUB_URL = "https://github.com";
```
AlemTuzlak, line 11: I left out the "/", not sure what your preference is on this one @brookslybrand

brookslybrand: Doesn't super matter to me. Thanks for doing this!

app/lib/tests/resources.server.test.ts
brookslybrand: Fun fact, having an integrated GitHub copilot for this was _super_ nice

image

brookslybrand (approved):
Awesome stuff! Thanks @AlemTuzlak

Issue #209, opened 2024-02-26
Upgrade Remix packages and use new Layout component


brookslybrand:
[Vite is stable](https://remix.run/blog/remix-vite-stable) 🎉 

Let's update our packages to the stable version.

Also, let's add the new [Layout export](https://remix.run/docs/en/main/file-conventions/root#layout-export) instead of the [Document](https://github.com/remix-run/remix-website/blob/main/app/root.tsx#L87-L139) component we leveraged instead.

If you want to take this on, feel free to split these 2 PRs if that's easier.

Kazuhiro-Mimaki:
I have a question about replacing new [Layout export](https://remix.run/docs/en/main/file-conventions/root#layout-export) instead of the [Document](https://github.com/remix-run/remix-website/blob/main/app/root.tsx#L87-L139) component.

Is it possible to pass some props like `forceDark`, `noIndex` from `App` / `ErrorBoundary` / `HydrateFallback ` to `Layout`?
It seems that `Layout` receive only `children` as the props, and I want to know how to pass other props.

brophdawg11:
IMO the new `Layout` component is syntactic sugar for _common_ use cases where the "shell" is identical between `Component`, `ErrorBoundary`, and/or `HydrateFallback`.  That doesn't mean that _every_ Remix app should definitely use a `Layout` component.  

I'm not familiar with everything that the `Document` component is doing (yet! I will dig in a bit when I have a sec) but if `App`/`ErrorBoundary` have to make enough decisions on their own that drive the `Document` via props - then composing via `Layout` may not be the right choice and maybe we just stick with `Document`.

Another option is whether or not `useRouteError` can be the thing we fork on to make those decisions in a Layout component.

PR #213, opened 2024-02-28
upgrade packages

Kazuhiro-Mimaki:
upgrade packages.

Related #209

### DIFF
--- a/data/posts/remix-heart-vite.md
+++ b/data/posts/remix-heart-vite.md
@@ -37,7 +37,7 @@ In fact, with Vite, Remix is no longer a compiler. **Remix itself is just a Vite
 
 ```ts
 // vite.config.ts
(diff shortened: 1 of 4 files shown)


brookslybrand (changes requested):
Looks good!

2.8.0 is about to be released as I'm writing this, so let's just go ahead and update to that version

Also, I realized that for some reason I started [pinning the Remix version](https://github.com/remix-run/remix-website/commit/d13246aa5fce9f9bf15269ebdc0735561342bdb6) a little while ago. Not sure why.

It ended up being important for Vite, since there were breaking changes as unstable Vite was being developed, but let's go ahead and add the `^` back to the beginning of all the Remix versions

Thanks @Kazuhiro-Mimaki!

brookslybrand (approved):

PR #216, opened 2024-02-29
chore: improve resource.server code to use zod

AlemTuzlak:
Changed the resource.server.ts code to use zod.


app/lib/resources.server.ts
```diff
@@ -5,40 +5,32 @@ import { processMarkdown } from "./md.server";
...
 
 export type CacheContext = { octokit: Octokit };
 
 const GITHUB_URL = "https://github.com";
 
-// TODO: parse this with zod
-let _resources: ResourceYamlData[] = yaml.parse(resourcesYamlFileContents);
+let _resources = yamlResourceSchema.parse(
+  yaml.parse(resourcesYamlFileContents),
+);
```
brookslybrand, line 17: Mind explicitly running this in a test? Probably in `resources.server.test.ts`

Right now we're inadvertently testing it, because in `resources.server.test.ts` we import `replaceRelativeLinks`, which ends up running this. However, if it fails it's going to be a bit confusing. Better to make it explicit with a test

And while we're here 🙈 let's put `resources.server.ts` and `resources.server.test.ts` in a directory together, just so they're a little more easily colocated

Something like `resources.server/index.ts` and `resources.server/resources.test.ts` should probably be good 🙏

brookslybrand (changes requested):
Looks really great! Just a suggestion about adding an explicit test and refactoring slightly

app/lib/resources.server.ts
AlemTuzlak: done

brookslybrand (approved):
Looks good, thanks!

PR #217, opened 2024-02-29
Fix meta tags

brookslybrand:
We had two broken meta tags in `meta.ts`

- `description` was not set up properly
- `twitter:card` was added to a `property` attribute instead of a `name` attribute. Seems like Twitter didn't care, but good to be consistent and correct


brookslybrand:
c.c @Kazuhiro-Mimaki just to keep you in the know

PR #220, opened 2024-02-29
fix meta description

Kazuhiro-Mimaki:
In #215 I made a mistake about meta description, and so fix up.


brookslybrand (approved):
Thanks for fixing these! I should have looked more carefully with my fix last week

PR #222, opened 2024-03-11
Single Fetch

brookslybrand:
Here's hoping this makes Fastly graphs go 📈


.npmrc
```diff
@@ -0,0 +1 @@
+force=true
```
brookslybrand, line 1: Don't want to leave this in long term, but makes it easier to deal with experimental releases here

app/entry.server.tsx
```diff
@@ -1,23 +1,152 @@
-import type { HandleDocumentRequestFunction } from "@remix-run/node";
+import { PassThrough } from "node:stream";
+
+import type { AppLoadContext, EntryContext } from "@remix-run/node";
+import { createReadableStreamFromReadable } from "@remix-run/node";
 import { RemixServer } from "@remix-run/react";
-import { renderToString } from "react-dom/server";
```
brookslybrand, line 3: Single fetch does not work with `renderToString` as @brophdawg11 pointed out. Went ahead and deleted these files and ran `npx remix reveal`. Let me know if there's any issue with that approach

brophdawg11 (changes requested):
Just dropping this here to block merging until I can get a final experimental cut - Jacob and I made a few final tweaks since the version this is pinned to

app/entry.server.tsx
```diff
@@ -1,23 +1,152 @@
...
+  request: Request,
+  responseStatusCode: number,
+  responseHeaders: Headers,
+  remixContext: EntryContext
+) {
+  return new Promise((resolve, reject) => {
+    let shellRendered = false;
+    const { pipe, abort } = renderToPipeableStream(
+      <RemixServer
+        context={remixContext}
```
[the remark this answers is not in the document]
jacob-ebey: We missed a type that's "shared" between runtimes somewhere. Package re-org can't come soon enough.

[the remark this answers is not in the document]
brophdawg11: huh - yeah I'll see what the error is - I thought we got the right fields added on both spots

app/entry.server.tsx
brophdawg11: Want to just keep them deleted if we don't do anything custom in them?

brophdawg11:
Updated to the latest experimental off `dev` in [a6325ce](https://github.com/remix-run/remix-website/pull/222/commits/a6325cef9bea7151afa2a10be31a7cd7699e1835)

app/entry.server.tsx
brophdawg11: I deleted them in my update - but feel free to add them back if you want them in!

app/entry.server.tsx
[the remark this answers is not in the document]
brophdawg11: Fixed - was the `renderMeta.streamCache`

brophdawg11 (approved):

PR #225, opened 2024-03-18
docs: fix casing Cloudflare

wouterds:
Minor issue but it grabbed my attention while re-reading this post, it's _Cloudflare_ not _CloudFlare_.


brookslybrand (approved):
Yep, good catch, thanks!

PR #226, opened 2024-03-19
Remove @remix-run/v1-meta

### LINKED ISSUE #214: Remove @remix-run/v1-meta
@Kazuhiro-Mimaki added a `getMeta` function in #196 which makes it much easier to add meta tags to various pages.

This function should also make it much easier to remove `@remix-run/v1-meta` usage and lean fully into the v2 convention

[There are 20 places we use the `metaV1`](https://github.com/search?q=repo%3Aremix-run%2Fremix-website+metaV1&type=code).

This issue can definitely be broken into multiple PRs, no need to try to migrate all of the places all at once. In fact, it will be much easier for me to review if it is broken up over a few PRs.

Be sure check that the meta tags look the same before and after the change, and that new tags are only introduced when it makes sense (for example, returning meta for a 404 probably shouldn't include `<meta name="twitter:card" content="summary_large_image">`

brookslybrand:
Ripping out the last bit of `@remix-run/v1-meta` usage. This PR does a couple of nice things:

- Removes `@remix-run/v1-meta` package
- Allows us to stop using `npm i --force` when testing experimental branches
- Improves the docs meta slightly by adding an image
- Fixed random typos

Big thanks to @Kazuhiro-Mimaki for creating the `getMeta` function and taking on the bulk of this work 🙏 

Resolves #214

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,0 @@
-force=true
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ FROM base as deps
 
 WORKDIR /remixapp
 
-ADD package.json package-lock.json .npmrc ./
+ADD package.json package-lock.json ./
 RUN npm install --include=dev
 
 # Setup production node_modules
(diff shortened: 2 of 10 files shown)


brookslybrand:
Whoops, goofed up and branched off of a branch with other commits. Will fix

Issue #233, opened 2024-03-21
Add auto scrolling to docs navigation


brookslybrand:
When you refresh or land on a certain page, it does not auto scroll/focus on the specific doc you're on in the left hand side, which is a bit disorienting

The "scroll" should be automatic (no smooth scrolling), in that that's what should be in view. This scroll resetting does not happen with the main content, so not totally sure what that's about.

https://github.com/remix-run/remix-website/assets/12396812/a5c3caae-ea87-42c6-ba1a-87d31dab7e4e

brookslybrand:
Can be done with #234

brookslybrand:
Thinking this is impossible without JS, I don't think anything but the main body holds scroll state between navigations/refreshes. The sidebar is a nav with `overflow-y: auto`

The JS fix is not worth it to me because you have to introduce CLS. Here's a slowed down network to exaggerate it. On a fast network it's just a really annoying flash

https://github.com/remix-run/remix-website/assets/12396812/0c25b83d-b065-4315-a8e1-bf7c3c0977a2

Closing, but if anyone can think of a solution let me know :)

PR #236, opened 2024-03-22
Docs Nav Scroll restoration

### LINKED ISSUE #234: Auto open current docs menu category
Once #232 is merged we'll have some cool collapsable nav menus!

Personally I think it'd be a bit more valuable if the menus were all automatically closed except for the category you're in. This is pretty easy to accomplish by breaking the `Menu` component apart and for each `li`/`details` component using something like the following:

```tsx
const isActivePath = useIsActivePath(category.slug);
return (
  <li
    key={category.attrs.title}
    className="[&:not(:last-child)]:mb-3"
  >
    <details
      className="group relative flex cursor-pointer flex-col"
      open={isActivePath}
    >
```

However, with this approach introduces some annoying UX where when you navigate out of the category it auto closes that category

https://github.com/remix-run/remix-website/assets/12396812/4393ed18-7e4d-4198-9bec-5145483a421d

I want both the default open category to be the matching one _and_ for navigations to keep currently open categories open. There are 2 ways to solve this that I can think of:

1. Sync open states to `localStorage`. Might be a bit heavy handed, but one benefit is that it will keep the opened categories open even between refreshes. However, it requires the introduction of a lot more state and likely `useEffect`. It also means that upon return visits the docs will have random categories open, which might be kind of confusing for a return user. As I'm writing this I don't like this idea
2. After hydration flip `open` back to `undefined` and let the browser handle it, that way client side navigations no longer effect the flow. Haven't tried it but hoping this is the easiest approach. Look to `useHydrated`

brookslybrand:
Improving the docs collapsable categories a bit

- Automatically open only current docs category -- resolves #234

### DIFF
--- a/app/routes/docs.$lang.$ref.tsx
+++ b/app/routes/docs.$lang.$ref.tsx
@@ -628,42 +628,75 @@ function NavMenuDesktop() {
 
 function Menu() {
   let { menu } = useLoaderData<typeof loader>();
+
   return menu ? (
     <nav>
       <ul>
-        {menu.map((category) => (
-          <li key={category.attrs.title} className="[&:not(:last-child)]:mb-3">
-            <details
-              className="group relative flex cursor-pointer flex-col"
-              open
-            >
-              <summary className="_no-triangle -mb-1 flex select-none items-center justify-between px-3 py-3 hover:bg-gray-50 active:bg-gray-100 dark:border-gray-700 dark:bg-gray-900 dark:hover:bg-gray-800 dark:active:bg-gray-700">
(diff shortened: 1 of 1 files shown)


brookslybrand:
I'm trying to get the scroll to automatically work (#233).

Tried a JS approach, but there's a pretty ugly CLS, which is not really worth it

```tsx
// in MenuLink component
      ref={(node) => {
        const isMounted = mounted.current;
        if (!isActive || isMounted) return;

        console.log("scroll into view", to);

        node?.scrollIntoView({
          behavior: "instant",
          block: "center",
        });
      }}
```

PR #241, opened 2024-03-29
Set `color-scheme` to `dark` when dark mode is enabled

AMEH64:
I updated the `tailwind.css` file to set `color-scheme` to `dark` when dark mode is enabled. This allows elements that do not have dark mode specific styles to fallback to the user-agent styles.

Below you will find a before and after image of the Remix docs loaded in the Brave browser. Focus your attention to the scrollbars, which are the only elements that I found to be impacted by this change.

Before:

image

After:

image

This is the same approach that the Tailwind CSS docs has taken: https://github.com/tailwindlabs/tailwindcss.com/blob/master/src/css/base.css

Kevin Powell has a great video that goes into more detail: https://www.youtube.com/watch?v=zFFuV_vXNhY

### DIFF
--- a/app/styles/tailwind.css
+++ b/app/styles/tailwind.css
@@ -38,6 +38,10 @@ html {
   font-size: 16px;
 }
 
+.dark {
+    color-scheme: dark;
+}
+
 [hidden] {
   display: none !important;
 }


brookslybrand:
This is really great @AMEH64, thank you for including this! Definitely makes those scrollbars look a lot better, and I'm sure it'll catch some other bad defaults

I went ahead and checked all the other pages, looks like everything looks as expected or better. Thanks again!

PR #242, opened 2024-03-31
Fix Overlapping Scroll in Docs Page Between Header and Elements

### LINKED ISSUE #239: Scroll on docs page has gross overlapping elements
Just see for yourself

https://github.com/remix-run/remix-website/assets/12396812/d06ee11c-0bb9-4ed9-a721-b22df316d40b

This just needs to be `position: sticky` instead of `position: fixed`. However, right now we leverage margins to push the main content to the right, which is kind of silly. The fix will require a bit of reconfiguring how the css here works

chialuka:
This PR fixes the overlap between elements when scrolling on the docs page. Here's how it looks now:

https://github.com/remix-run/remix-website/assets/42539315/ae9e2e74-fbbf-4d7d-920c-5d098332eb57

Closes #239

### DIFF
--- a/app/routes/docs.$lang.$ref.$.tsx
+++ b/app/routes/docs.$lang.$ref.$.tsx
@@ -158,7 +158,7 @@ export default function DocPage() {
 
 function LargeOnThisPage({ doc }: { doc: SerializeFrom<Doc> }) {
   return (
-    <div className="sticky top-36 order-1 mt-20 hidden max-h-[calc(100vh-9rem)] w-56 flex-shrink-0 self-start overflow-y-auto pb-10 xl:block">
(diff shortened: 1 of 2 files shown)


brookslybrand:
Hey thanks so much @chialuka for opening up a PR for this

Unfortunately this ties the side nav and the main body to the same scroll, when they really should have 2 separate scroll areas

Uploading scroll_bug.mov…

brookslybrand:
@chialuka I went ahead and implemented this in #253. Thank you so much for taking a stab at it! This was a useful reference as I worked on a solution that hit all the corner-cases I was concerned about

chialuka:
> @chialuka I went ahead and implemented this in #253. Thank you so much for taking a stab at it! This was a useful reference as I worked on a solution that hit all the corner-cases I was concerned about

@brookslybrand, I'm so sorry I got tied down with work and couldn't get to this. I'll take a stab at another PR and hope I can contribute to the project

PR #244, opened 2024-04-02
chore(resources): add @mcansh/vite-svg-sprite-plugin

mcansh:
@mcansh/vite-svg-sprite-plugin is a vite plugin that will transform any imported svg files and combine them into an svg sprite sheet


brookslybrand:
Whoa, that seems super useful. Thanks @mcansh!

Issue #243, opened 2024-04-01
Mermaid Support


brophdawg11:
It would be awesome to add mermaid support in our markdown files via `mermaid` code blocks like github has:

https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/

Then we could include all sorts of [diagrams](https://mermaid.live/edit#pako:eNpFz70OgjAQAOBXaW4W2WvCYCQ6yAST6dLQQ0hoi9fWnxDe3QqoN90l3_2NUFuFwMHhLaCp8dDJK0ktDIuxJ_twSCzJMlYi3ZE4O-YVa70fHE9TQt09txTMbvGLSSJfOzk7VcVZGNiARtKyU3HV-MECfIsaBfCYKmxk6L0AYaZIw6Ckx1x13hLwRvYONyCDt-XL1MA9Bfyi9dyfGqS5WPuvcR5SLD_Or05vpOdSKw) in our markdown files.

References:
 - https://github.com/mermaid-js/mermaid
 - https://github.com/remcohaszing/rehype-mermaid

kandros:
👋 @brophdawg11 
I was able to make it work
<img width="2282" alt="image" src="https://github.com/remix-run/remix-website/assets/4562878/d89f451a-ee2e-40a9-b659-23aab3c41c87">

Turns out rehype-mermaid depends on [playwright](https://playwright.dev/) in order to render the image, do you think this is acceptable or is a blocker to continue?

brophdawg11:
I think folks on twitter were saying mermaid adds pretty significant [bundle bloat](https://x.com/imagineherhappy/status/1774866341981475321) - can you confirm/deny the change in client-side JS payload when this is added?

If there is a bloat I think we may need to look into lazily loading the mermaid dependencies and rendering the diagrams client side only - although that could introduce a potentially bad CLS if the diagram was above the fold.  I wonder if they could be pre-rendered to SVG's on the server somehow

PR #250, opened 2024-04-04
log error when doc generation fail

kandros:
I had a hard time working on #243 due to missing logs.

This made understanding what was going easier


brookslybrand:
I ran into this too but was too lazy to fix, thank you @kandros!

I'll merge this and #248 tomorrow. We're currently investigating a memory leak. We believe we've caught it, but want to avoid touching the site until tomorrow to make sure.

brookslybrand:
Thanks for looking into #243, that one seems pretty tricky but @brophdawg11 will be very happy if you crack it

brookslybrand (approved):

PR #253, opened 2024-04-17
Make doc nav sticky instead of fixed

### LINKED ISSUE #239: Scroll on docs page has gross overlapping elements

brookslybrand:
Fix currently deployed to [staging](https://remix.run/docs/en/main/start/changelog)

https://github.com/remix-run/remix-website/assets/12396812/b7c466c4-ee58-4a5e-8479-8da13d9f5b77

Fixes #239

### DIFF
--- a/app/routes/docs.$lang.$ref.tsx
+++ b/app/routes/docs.$lang.$ref.tsx
@@ -119,24 +119,27 @@ export default function DocsLayout() {
         }
       >
         <InnerContainer>
-          <NavMenuDesktop />
-          <div
-            ref={docsContainer}
-            className={cx(
-              // add scroll margin to focused elements so that they aren't
-              // obscured by the sticky header
-              "[&_*:focus]:scroll-mt-[8rem] lg:[&_*:focus]:scroll-mt-[5rem]",
-              "min-h-[80vh]",
-              "lg:ml-72 lg:pl-6 xl:pl-10 2xl:pl-12",
-              !changingVersions && navigating
-                ? "opacity-25 transition-opacity delay-300"
-                : "",
-            )}
-          >
-            <Outlet />
-          </div>
-          <div className="min-w-0 pt-8 sm:pt-10 lg:ml-72 lg:pl-6 lg:pt-12 xl:pl-10 2xl:pl-12">
-            <Footer />
+          <div className="block lg:flex">
+            <NavMenuDesktop />
+            <div
+              ref={docsContainer}
+              className={cx(
+                // add scroll margin to focused elements so that they aren't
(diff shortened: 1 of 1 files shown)


app/routes/docs.$lang.$ref.tsx
```diff
@@ -119,24 +119,27 @@ export default function DocsLayout() {
...
+          <div className="block lg:flex">
+            <NavMenuDesktop />
+            <div
+              ref={docsContainer}
+              className={cx(
+                // add scroll margin to focused elements so that they aren't
+                // obscured by the sticky header
+                "[&_*:focus]:scroll-mt-[8rem] lg:[&_*:focus]:scroll-mt-[5rem]",
+                // Account for the left navbar
+                "min-h-[80vh] lg:ml-3 lg:w-[calc(100%-theme(spacing.72))]",
```
silvenon, line 131: Seems like you don't need width here, one less magic number to keep track of.
```suggestion
                "min-h-[80vh] lg:ml-3 lg:flex-1",
```

silvenon (commented):
Looks good on all sizes 👍

If you want to completely avoid duplicating values, you can set header height in a CSS variables via classes. On a common container you set the class `[--header-height:theme(spacing.16)]`, then `h-[--header-height]` and `h-[calc(100vh-var(--header-height))]`.

brookslybrand:
> If you want to completely avoid duplicating values, you can set header height in a CSS variables via classes. On a common container you set the class `[--header-height:theme(spacing.16)]`, then `h-[--header-height]` and `h-[calc(100vh-var(--header-height))]`.

Great suggestion! I'll have to make these adjustments when I get back. Was trying to squeeze this in before going out of town, but I don't think I gave myself enough time.

Thanks so much for the review!

app/routes/docs.$lang.$ref.tsx
brookslybrand: Just kidding, I'm back and trying to get this in real quick

I definitely love the idea of not having another magic number! However, it seems like you can get into a weird state here with a squished left nav.

image

If I make the left nave have a `min-w-72`, then at this size some of the doc gets cut off and you can't even horizontally scroll to it.

I think I'm just going to take the same approach with the height that you shared, but if you have a way to account for all these issues I'm all ears!

PR #261, opened 2024-05-13
refactor: replace tailwind-extras.html with safelist feature

kisaragi-hiu:
> this is silly, but our docs have some tailwind classes so I've copied them here so that in CI tailwind will add the right classes, gonna think of something better but I'm really tired and just want to ship this ...

Tailwind has [safelist](https://tailwindcss.com/docs/content-configuration#safelisting-classes) for this purpose (forcing classes to be generated no matter what) and can be used instead of a quick workaround. Hopefully this is something better.

I rewrote the comment like this:

> docs have some tailwind classes, tell tailwind to always add these classes so the right classes are still added in CI

which should hopefully still adequately explain the intent of safelisting these classes.

### DIFF
--- a/tailwind.config.mjs
+++ b/tailwind.config.mjs
@@ -7,7 +7,20 @@ const defaultTheme = require("tailwindcss/defaultTheme");
 /** @type {TailwindConfig} */
 export default {
   mode: "jit",
-  content: ["./app/**/*.{ts,tsx}", "./data/**/*.md", "./tailwind-extras.html"],
+  content: ["./app/**/*.{ts,tsx}", "./data/**/*.md"],
+  // docs have some tailwind classes, tell tailwind to always add these classes
+  // so the right classes are still added in CI
+  safelist: [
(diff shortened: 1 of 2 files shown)


tailwind.config.mjs
```diff
@@ -7,7 +7,20 @@ const defaultTheme = require("tailwindcss/defaultTheme");
 /** @type {TailwindConfig} */
 export default {
   mode: "jit",
-  content: ["./app/**/*.{ts,tsx}", "./data/**/*.md", "./tailwind-extras.html"],
+  content: ["./app/**/*.{ts,tsx}", "./data/**/*.md"],
```
brookslybrand, line 10: Thanks for this solution!

I don't actually know if we need this at all, since we're already looking at `.md` in our `content`

I think we can remove the `safelist` as well and go ahead and merge this in

brookslybrand:
Resolved in #276

Thanks for pointing this issue out!

PR #276, opened 2024-06-17
Misc Tailwind updates

ryanleichty:
Miscellaneous updates related to Tailwind.

- Updated `tailwindcss`.
- Updated `prettier-plugin-tailwindcss`.
- Ran the `format` script. `prettier-plugin-tailwindcss` v6 added the functionality of removing extra spaces and duplicate classes which is a great quality of life improvement. There was one spot in the footer where the spaces removed would've caused an issue because `clsx` wasn't being used to combine classes so I updated that.
- Move Tailwind classes from Ryan’s extra HTML file to the Tailwind config safelist.
- Remove display font family. With the removal of Founders Grotesk and using Inter for everything, there's no need for having the additional `display` font family. Removed references to it to keep things clean.
- Remove jit mode in the config as it's enabled by default now.
- Remove `@tailwindcss/aspect-ratio` plugin as the `aspect-ratio` property is widely supported now.


brookslybrand:
Thank you @ryanleichty for this cleanup! I should be able to run through and get it merged this afternoon 🤘

brookslybrand (approved):
Lovely, thanks so much!

PR #277, opened 2024-06-17
Update OG image

ryanleichty:
I meant to update this back when I swapped out Founders Grotesk with Inter. It seems that you have to load each font weight separately versus using the variable font weights. This primarily updates the text that was previously Founders Grotesk bold to Inter black. The font files I uploaded are a subset of Inter using basic latin characters ([created via this tool](https://fonttools.readthedocs.io/en/latest/subset/)).

I also tweaked the size and alignment of elements and added the logo as a separate element instead of being included in the background image.

### Single author
CleanShot 2024-06-17 at 11 09 18@2x
CleanShot 2024-06-17 at 13 58 10@2x

### Multiple authors
CleanShot 2024-06-17 at 14 00 07@2x
CleanShot 2024-06-17 at 14 00 27@2x


brookslybrand:
Lovely, thank you for doing this!

PR #278, opened 2024-06-18
Fix invisible text

Nautman:
There is invisible text (i.e. text that is white, same as background) when showing a subscription confirmation.
image

The fix looks like this:
image


brookslybrand (approved):
Whoops! Thanks for the fix @Nautman

Issue #284, opened 2024-07-10
Remove background from brand page assets


brookslybrand:
When we setup the brand page we copied the assets from [this Google Drive](https://drive.google.com/drive/u/0/folders/1pbHnJqg8Y1ATs0Oi8gARH7wccJGv4I2c). Those assets have backgrounds attached them, which is not ideal when people are trying to use our logo for a presentation, website, etc.

https://github.com/remix-run/remix-website/assets/12396812/b2895cc5-b1f8-4d1c-8693-3e88f1bea601

PR #285, opened 2024-07-10
Add transparency to logo images on the brand assets page

tquirino:
The current PNG files don't have a transparent background. This makes it more difficult to use the Remix logo. I've re-exported each image to fix this, so that they all have transparent backgrounds.


brookslybrand:
Sweet! Will you delete the old files and update the name of each of the new files? That will allow users to download the correct asset from the [brand page](https://remix.run/brand)

image

Also, will you remove the background from each of the SVGs? Those might be easier to just re-export from the figma file

brookslybrand (approved):
Awesome! Thank you @tquirino

Resolves #284

PR #293, opened 2024-08-19
Use build-time og images

kettanaito:
## Changes

- Adopts [remix-og-image](https://github.com/kettanaito/remix-og-image) Vite plugin to generate OG images for blog posts at build-time.

## Why?

The plugin has a nice [summary of its benefits](https://github.com/kettanaito/remix-og-image?tab=readme-ov-file#features). Also summarizing here:

- In-browser rendering. No more loading fonts, styling via `style`, dealing with the limitations of runtime-based solutions. 
- OG images are preview-able while developing because an OG image is now a regular UI Remix route: `/og/blog/:slug` in the browser to iterate on the image. Debuggable with your favorite tools. 
- No runtime cost for OG images. 
- Also using `*.webp` for OG images as it yields to ~61% smaller file size with superior quality to JPEG.

## Examples

I tried to re-create what you have with Satori but using Tailwind. Sharing some of the generated OG images below.

> GitHub doesn't support uploading WEBP. The images below are JPEG, so sorry for poor quality. Please see them for yourself by running `npm run build`. The images will be in `/public/og`. 

### Post with unique background

fog-of-war

### Post without background (default background)

react-router-v6

### Post with multiple authors

remix-heart-vite

### Post with a long title

incremental-path-to-react-19

## Roadmap

- [ ] Remove `satori` and the previous SVG-based runtime OG generation setup (once the changes are confirmed as good, will remove that). Looking forward to that nice diff!

### DIFF
--- a/app/routes/og.blog.$slug.tsx
+++ b/app/routes/og.blog.$slug.tsx
@@ -1,0 +1,150 @@
+import invariant from "tiny-invariant";
+import getEmojiRegex from "emoji-regex";
+import { type LoaderFunctionArgs } from "@remix-run/node";
+import { useLoaderData } from "@remix-run/react";
+import { isOpenGraphImageRequest, OpenGraphImageData } from "remix-og-image";
+import {
+  BlogAuthor,
+  getBlogPost,
+  getBlogPostListings,
+} from "~/lib/blog.server";
+import { Wordmark } from "~/ui/logo";
+
+interface AuthorWithImage extends BlogAuthor {
+  imgSrc: string;
+}
+
+export async function openGraphImage() {
+  const posts = await getBlogPostListings();
+
+  return posts.map<OpenGraphImageData>((post) => {
+    return {
+      name: post.slug,
+      params: { slug: post.slug },
+    };
+  });
+}
+
+export async function loader({ request, params }: LoaderFunctionArgs) {
+  if (isOpenGraphImageRequest(request)) {
+    return openGraphImage();
+  }
+
+  let { slug } = params;
+  invariant(!!slug, "Expected slug param");
+
+  let requestUrl = new URL(request.url);
+  let siteUrl = requestUrl.protocol + "//" + requestUrl.host;
+  let post = await getBlogPost(slug);
+
+  let backgroundImage: string | undefined;
+
+  if (post.ogImage) {
+    backgroundImage = `url(${siteUrl}${post.ogImage})`;
+  } else {
+    let socialBackground = await import(
+      "./img.$slug/social-background.png?arraybuffer"
+    );
+    let base64 = arrayBufferToBase64(socialBackground.default);
+    backgroundImage = `url("data:image/png;base64,${base64}")`;
+  }
+
+  return {
+    post: {
+      ...post,
+      dateDisplay: stripEmojis(post.dateDisplay),
+      authors: post.authors.map((author) => {
+        return {
+          ...author,
+          title: stripEmojis(author.title),
+          imgSrc: getAuthorImgSrc(siteUrl, author.name),
+        };
+      }),
+    },
+    backgroundImage,
+  };
+}
+
+export default function OpenGraphImageTemplate() {
+  const { post, backgroundImage } = useLoaderData<typeof loader>();
+
+  return (
+    <div
+      id="og-image"
+      className="flex h-[628px] w-[1200px] items-end justify-center border border-gray-800 bg-no-repeat"
+      style={{ backgroundImage, backgroundSize: "100% 100%" }}
+    >
+      <div className="flex h-full w-full flex-col justify-between p-16 text-2xl text-white">
+        <div className="flex flex-col">
+          <div className="text-gray-200">{post.dateDisplay}</div>
+          <h1 className="mt-6 text-pretty pr-40 text-7xl font-black leading-[1.1]">
+            {post.title}
+          </h1>
+        </div>
+
+        <Authors authors={post.authors} />
+      </div>
+    </div>
+  );
+}
+
+// Taken from https://stackoverflow.com/questions/9267899/arraybuffer-to-base64-encoded-string
+function arrayBufferToBase64(buffer: ArrayBuffer) {
+  let binary = "";
+  let bytes = new Uint8Array(buffer);
+  let len = bytes.byteLength;
+  for (let i = 0; i < len; i++) {
+    binary += String.fromCharCode(bytes[i]);
+  }
+  return btoa(binary);
+}
+
+function getAuthorImgSrc(siteUrl: string, name: string) {
+  let authorNameSlug = name
+    .toLowerCase()
+    .replace(/[^a-z0-9\s_-]/g, "")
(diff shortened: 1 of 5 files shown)


app/routes/og.blog.$slug.tsx
```diff
@@ -0,0 +1,150 @@
...
+  getBlogPost,
+  getBlogPostListings,
+} from "~/lib/blog.server";
+import { Wordmark } from "~/ui/logo";
+
+interface AuthorWithImage extends BlogAuthor {
+  imgSrc: string;
+}
+
+export async function openGraphImage() {
```
kettanaito, line 17: This feels like an extremely Remix way of doing this. Looking forward to hearing your thoughts on this.

kettanaito:
## Deployment

A single requirement this plugin has is for your build to run a Puppeteer binary. You seem to be using Fly, which I believe is pretty flexible in terms of Docker images you use to deploy your app. 

Can someone please double-check if you have Chromium installed already? If not, we should add that to the image to skip downloading it on every build. cc @brookslybrand 

> If this would be a complete no-go, you can also upload the generated images manually anywhere (e..g a CDN) using the [`writeImage`](https://github.com/kettanaito/remix-og-image?tab=readme-ov-file#what-if-i-dont-want-to-write-images-to-disk) option. It gives you the generated image's buffer.

kettanaito:
I think with this linting rules, you may want to enable [`verbatimModuleSyntax`](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax) option in TypeScript. It will force `type` imports on the transpiler level. 

```
Warning:   5:1  warning  Import "OpenGraphImageData" is only used as types  @typescript-eslint/consistent-type-imports
```

kentcdodds:
This is neat! The problem with this approach is that it requires a rebuild of the whole site when new posts are published or a change is made to the og images. Additionally, the build times increases linearly with the number of pages which can aggravate the rebuild issues. Taking this to the logical conclusion is why Gatsby and Next.js came up with Deferred Static Generation (DSG) and Incremental Static Regeneration (ISR).

Having things be generated at runtime is a desirable feature for these reasons.

kettanaito:
Thanks for the feedback, @kentcdodds! Those are valid concerns, allow me to address some and share my opinion on others. 

## Changes require a rebuild

This will be my most controversial opinion, perhaps, but I truly believe that runtime content rendering, while it has a place to be, is a bad idea for most people. In practice, blogs, portfolios, and personal websites are static by nature. Pushing the content rendering to runtime adds unnecessary complexity and fragility (what if my runtime function pulls from a malformatted MDX? The site is down). 

I understand that may be desired in the case of highly-dynamic website, like yours. Realistically, very few people build or even need that level of detail. In fact, when I created a site for my wife using Remix, I quickly learned that it was a mistake to use it for a de-facto static website (it's still one of my painpoints with the framework). It introduces a degree more issues compared to the value it provides (I have to use verbose runtime MDX rendering, Remix doesn't support relative assets like images so I need to create explicit resource routes for those, etc). 

**That being said, the point itself is valid**. Having to wait for an entire build to finish just to see an update to an OG image is likely too much. I think the solution here is asset caching during the build. `remix-og-image` already has caching built-in. It won't generate anything if there's been no changes to the relevant content routes. I need to see how to make its cache persist across builds, and then you will only wait on generating the stuff that actually changed. I believe that would address your concern. 

## Linear build time increase

I agree on this one too. I mean, it's x10 build time increase as of now, that's not nice! 

I have at least four things to try to improve the plugin's performance:

1. Spawn the browser earlier (right now the plugin waits until the end of the build).
2. Spawn the Vite preview server earlier (the same story).
3. Fetch all OG routes' `loader` data in a single fetch using `_routes` (right now, the plugin does a request for each route individually). This will only affect Remix apps with the Single fetch enabled (while it's experimental anyway).
4. Optimize the way it spawns pages in Puppeteer.

I believe implementing these will yield to a decrease in build time. 

It's worth mentioning that build-time generation _will take time_ by design. As long as that time is appropriate, I'd argue it's a fantastic price to pay for all the benefits that the build-time generation provides, proper rendering and easier setup being just a few. If you consider how much compute this solves for your app in the long run, it's more than worth it. 

## DSG and ISR

Both are fantastic techniques that solve real problems. At the same time, I must admit, I've never faced any of those problems building websites and publishing content for almost a decade now. I can safely say most people haven't either. 

The majority of folks don't have millions of pages. Not even hundreds. It's good if they have a dozen! Transforming MDX is really fast (that's one of the reasons you can put it on runtime, to begin with). 

**There is an important exception**. If your content is dynamic, as in something on the OG image changes independently from your post (like GitHub's PR title on the PR OG image preview), **then this plugin is not for you**. Then, the runtime generation is indeed a solution to a problem here. But it's not in all the other cases, which I believe is the majority (like this website!). 

---

Your feedback comes down to performance and optimization, and I'm happy about that! The plugin is extremely raw, and trying it out on the Remix's site gave me a lot of hopes in its future. The fact that I was able to re-create your setup being entirely unfamiliar with the code base and the previous setup in under an hour is incredible. I spent an hour tweaking Satori just to understand it forces me to load a TTF version of my font that is x3 larger just so I could generate an image. That doesn't sound like a good design to me. 

> I have to emphasize I don't see this plugin as a panacea. Neither do I see the runtime image generation as one. Those are two different techniques to solve different problems. It's bad when you employ a tool to solve problems you didn't have, to begin with. This applies to both of them. This plugin won't always make sense, and I will make it abundantly clear in its repo what it's for. 

So, I will keep working on the plugin's performance and let's see where it goes!

kiliman:
When I was building out my blog, which ran on Cloudflare Workers, I decided that I wanted my MDX content to be compiled at "build" time. My content was static files committed to the repo, so I had a script using GitHub Actions that compiled the MDX and pushed the results to the Worker KV store. This way, there was no runtime cost. It even included Etags, so I didn't even need to send the result if it was already cached.

For local development, I created a content watcher that would compile and store in local KV whenever I saved the MDX file.

This resulted in a very nice workflow that optimized for reading as well as change.

kettanaito:
- Improving CI support: https://github.com/kettanaito/remix-og-image/pull/13
- Performance effort: https://github.com/kettanaito/remix-og-image/pull/14

kettanaito:
I am looking at the Fly Docker image to install Chromium, allowing this plugin to run (the default image doesn't have it installed). Will try to push the changes later next week. 

> Already have this plugin confirmed working on Cloudflare, expect Vercel to be the same. Not sure why Chromium is not installed as a part of `puppeteer` on Fly.

kettanaito:
## Updates

With https://github.com/kettanaito/remix-og-image/releases/tag/v0.3.5 and preceding releases, I've shaved about 3s from each generated image. I've also added proper caching on CI, which should be supported in Fly as well. This should improve the build time and produce dramatically faster incremental builds. 

I've also added `browser.executablePath` option since on Fly it looks like we'd have to install Chromium and provide an exec path to it manually. Tweaking the existing Docker config to do that...

PR #294, opened 2024-08-30
Draft: Testing out Single Fetch

brookslybrand:
Testing out the latest [Remix nightly release](https://github.com/remix-run/remix/releases/tag/v0.0.0-nightly-24d1b2186-20240830) and running into a few problems

## Optimize Deps unexpected error

When `npm run dev`, I get the following error, and I don't get it on the previous nightly

```sh
Re-optimizing dependencies because lockfile has changed
Express server listening on port 3000 (http://localhost:3000)
✘ [ERROR] Failed to resolve entry for package "jimp". The package may have incorrect main/module/exports specified in its package.json. [plugin vite:dep-pre-bundle]

    node_modules/vite/node_modules/esbuild/lib/main.js:1373:21:
      1373 │         let result = await callback({
           ╵                      ^
```

It's easy to solve by adding the following to the Vite config

```ts
// vite.conf.mts

optimizeDeps: {
	exclude: ["svg2img"],
},
```

However, this took me a second to figure out, and mostly because I knew we'd made some changes to vite dependency optimization stuff. I'm nervous this is going to cause people's apps to break in dev unexpectedly.

## Single Fetch types

With the single fetch types, there seem to be 2 issues

1. The inability to use `interface` still seems to be a problem. Is this expected? I thought this was something that we could resolve with the plugin to `turbo-stream`
2. The return type seems to be wrong, it thinks it's returning an object with a `data` key

I recorded a video walking through these issues

https://github.com/user-attachments/assets/7835126c-3066-4be4-805b-4ffc73435a90

### DIFF
--- a/app/lib/blog.server.ts
+++ b/app/lib/blog.server.ts
@@ -159,12 +159,12 @@ interface MarkdownPost {
   html: string;
 }
 
-export interface BlogAuthor {
+export type BlogAuthor = {
   name: string;
   title: string;
   avatar: string;
-}
+};
 
-export interface BlogPost extends Omit<MarkdownPost, "authors"> {
+export type BlogPost = Omit<MarkdownPost, "authors"> & {
   authors: BlogAuthor[];
-}
+};
--- a/app/routes/_extras.blog.$slug.tsx
+++ b/app/routes/_extras.blog.$slug.tsx
@@ -5,7 +5,7 @@ import type {
 } from "@remix-run/node";
 import { useLoaderData } from "@remix-run/react";
 import type { MetaFunction } from "@remix-run/react";
-import { json } from "@remix-run/node";
+import { unstable_data } from "@remix-run/node";
 import invariant from "tiny-invariant";
 
 import { getBlogPost } from "~/lib/blog.server";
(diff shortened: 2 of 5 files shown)


pcattori:
@brookslybrand something is definitely not working correctly since when you hovered `post` you were getting `JsonifyObject` stuff when with single-fetch types, you shouldn't be getting that at all

Screenshot 2024-08-30 at 10 46 26 AM

pcattori:
For the issue with optimize dependencies, it looks like `jimp` is just configured incorrectly (https://publint.dev/jimp@0.22.12). You would have hit this issue before automatic dependency optimization if you ever hit the URL for the resource route that uses `svg2img` i.e. `img.$slug`

brookslybrand:
> For the issue with optimize dependencies, it looks like `jimp` is just configured incorrectly (https://publint.dev/jimp@0.22.12). You would have hit this issue before automatic dependency optimization if you ever hit the URL for the resource route that uses `svg2img` i.e. `img.$slug`.

Interesting... I've gone directly to that url when testing our og images and it's never failed before I don't think.

> We can definitely add some troubleshooting docs for this. Also is the `Failed to resolve entry for package` blocking (does it crash the dev server and not let you dev)?

Yep, completely crashed it, couldn't start the dev server

It makes sense _why_ this happened, I'm just a little nervous about this happening to people on a patch release

pcattori:
Oh your `tsconfig.json` is configured to only include `**/*.ts` and `**/*.tsx`, but your Vite config has a `.mts` extension. That means that code in `.mts` files doesn't get used as part of the projects types

I don't think `.mts` is meaningful in this context, so should probably just rename to `vite.config.ts`

PR #295, opened 2024-09-05
Update to Remix 2.12.0

brookslybrand:


app/routes/[_]actions.newsletter.tsx
```diff
@@ -9,14 +9,14 @@ export const action = async ({ request }: ActionFunctionArgs) => {
...
 
   try {
     await subscribeToNewsletter(email);
   } catch (e: any) {
-    return json({ error: e.message || "Unknown error", ok: false });
+    return data({ error: e.message || "Unknown error", ok: false });
   }
 
-  return json({ error: null, ok: true });
+  return data({ error: null, ok: true });
```
Comment (brophdawg11) [inline app/routes/[_]actions.newsletter.tsx:21]: This and the prior one don't _need_ `data()` because they don't care about the `status` but there's a known issue in our type inference that doesn't like the mixing and matching of data/json/etc. and raw objects so using data everywhere to avoid that.

app/routes/docs.$lang.$ref.$.tsx
```diff
@@ -156,7 +152,7 @@ export default function DocPage() {
   );
 }
 
-function LargeOnThisPage({ doc }: { doc: SerializeFrom<Doc> }) {
+function LargeOnThisPage({ doc }: { doc: Doc }) {
```
brophdawg11, line 155: With Single Fetch now we just get back a `Doc`

brophdawg11 (approved):

app/routes/[_]actions.newsletter.tsx
Comment (brookslybrand) [reply app/routes/[_]actions.newsletter.tsx:21]: I left a comment on the [RR PR](https://github.com/remix-run/react-router-website/pull/118#discussion_r1757522208), should have read this one first.

Thanks for the explanation!

PR #297, opened 2024-09-18
Drop `json` utility usage

brookslybrand:


app/routes/$.tsx
```diff
@@ -44,7 +44,7 @@ function handleStaticFileRequests(param: string | undefined) {
   if (
     SAFE_STATIC_FILE_EXTENSIONS.some((ext) => !!(param && param.endsWith(ext)))
   ) {
-    throw json({}, { status: 404 });
+    throw new Response(null, { status: 404 });
```
brophdawg11, line 47: Out of curiosity does this work if you `throw unstable_data(null, { status: 404 })`?  It should so if not we should fix that.  You shouldn't need to use `new Response` outside of a resource route

app/routes/[_]actions.newsletter.tsx
```diff
@@ -15,8 +15,8 @@ export const action = async ({ request }: ActionFunctionArgs) => {
   try {
     await subscribeToNewsletter(email);
   } catch (e: any) {
-    return data({ error: e.message || "Unknown error", ok: false });
+    return { error: e.message || "Unknown error", ok: false };
```
Comment (brophdawg11) [inline app/routes/[_]actions.newsletter.tsx:18]: :chefs-kiss:

app/routes/$.tsx
brookslybrand: Good question, yep that works perfectly fine. It's just fun to throw responses 😉 

On the note of a resource route, if you do throw `unstable_data` you get an error like the following. It might be nice to provide a little bit more clear message, since I imagine this will trip some people up if they don't totally grok what `data` is doing/what resource routes are

```
{
  "message": "Unexpected Server Error",
  "stack": "Error: Unexpected Server Error\n    at errorResponseToJson (/Users/brookslybrand/code/remix-website/node_modules/@remix-run/server-runtime/dist/server.js:438:26)\n    at handleResourceRequest (/Users/brookslybrand/code/remix-website/node_modules/@remix-run/server-runtime/dist/server.js:429:14)\n    at requestHandler (/Users/brookslybrand/code/remix-website/node_modules/@remix-run/server-runtime/dist/server.js:156:18)\n    at /Users/brookslybrand/code/remix-website/node_modules/@remix-run/express/dist/server.js:41:22"
}
```

brophdawg11: ok yeah we can make that more clear and potentially handle simple data values and convert them to response son our end if we think that's a good idea

brophdawg11 (approved):

PR #306, opened 2024-11-05
Fix URL: fog-of-war => lazy-route-discovery

bewildergeist:
The first link in [Matt's blogpost](https://remix.run/blog/fog-of-war) is [currently 404'ing](https://remix.run/docs/en/main/guides/fog-of-war). This PR:

- Updates the link to point to the ["Lazy Route Discovery" guide in the docs](https://remix.run/docs/en/main/guides/lazy-route-discovery)
- Adds a redirect from the `fog-of-war` URL to the `lazy-route-discovery` URL. I appended it to the bottom of the file, but perhaps it should be moved elsewhere? Or maybe not necessary at all? 🤷‍♂️

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -62,3 +62,4 @@
 /docs/future/server-bundles           /docs/guides/server-bundles          302
 /docs/future/spa-mode                 /docs/guides/spa-mode                302
 /docs/future/vite                     /docs/guides/vite                    302
+/docs/guides/fog-of-war               /docs/guides/lazy-route-discovery    302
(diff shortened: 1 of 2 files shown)


machour (approved):
Thanks for catching this!

_redirects
```diff
@@ -62,3 +62,4 @@
 /docs/future/server-bundles           /docs/guides/server-bundles          302
 /docs/future/spa-mode                 /docs/guides/spa-mode                302
 /docs/future/vite                     /docs/guides/vite                    302
+/docs/guides/fog-of-war               /docs/guides/lazy-route-discovery    302
```
brookslybrand, line 65: Given that this was always a 404 I don't think we need to setup a custom redirect for it

```suggestion

```

brookslybrand (approved):
Thanks @bewildergeist! Can't believe we hadn't noticed this yet 🤦

PR #320, opened 2025-02-27
Add Split Route Modules blog post

markdalgleish:


data/posts/split-route-modules.md
```diff
@@ -0,0 +1,137 @@
...
+date: 2025-02-27
+image: /blog-images/headers/react-router-v7.jpg
+ogImage: /blog-images/headers/react-router-v7.jpg
+imageAlt: "The React Router logo"
+imageDisableOverlay: true
+authors:
+  - Mark Dalgleish
+---
+
+With the release of [React Router v7.2.0](https://github.com/remix-run/react-router/releases/tag/react-router%407.2.0), we’ve introduced a new opt-in framework feature called Split Route Modules. In this post we’ll explore the performance problem that Split Route Modules solves, how it works, and how to use it today.
```
brophdawg11, line 13: We should probably call out that it's currently unstable and not for production use?  Or are we planning to publish this once it's stabilized?

data/posts/split-route-modules.md
```diff
@@ -0,0 +1,137 @@
...
+Here we have a minimal `clientLoader` that makes a basic fetch call to an external API. The route component, on the other hand, is much larger.
+
+Unfortunately, because these exports are both contained within the same module, the client loader has to wait for the rest of the route module to download before it can start running. If the client loader needs to perform an async operation like hitting an external API, this delay can be felt by the user.
+
+To visualize this as a timeline:
+
+```
+Get Route Module:  |--=======|
+Run clientLoader:            |-----|
+Render:                            |-|
```
brophdawg11, line 44: I used excalidraw for my [route.lazy](https://remix.run/blog/lazy-loading-routes#render--fetch-chains) blog post diagrams if we want to spruce these up and use something familiar from the blog.  I may have saved off the file too - let me know if you want me to look around for it.

data/posts/split-route-modules.md
```diff
@@ -0,0 +1,137 @@
...
+
+This would instead be optimized to the following:
+
+```
+Get clientLoader:     |--|
+Get clientAction:     |~~~|
+Get HydrateFallback:  SKIPPED
+Get Component:        |=======|
+Run clientLoader:        |-----|
+Render:                        |-|
```
brophdawg11, line 104: Is this missing a `Run clientAction` step?

data/posts/split-route-modules.md
```diff
@@ -0,0 +1,137 @@
...
+import type { Config } from "@react-router/dev/config";
+
+export default {
+  future: {
+    unstable_splitRouteModules: true,
+  },
+} satisfies Config;
+```
+
+If your project is especially performance sensitive, you can set `future.unstable_splitRouteModules` to `"enforce"`. This will break the build if any route module cannot be split.
```
brophdawg11, line 124: This feels like it's missing some context - should we include a small section with the scenarios that prohibit splitting and maybe a small code example?

brophdawg11 (commented):
Great write up!

data/posts/split-route-modules.md
brookslybrand, line 13:
```suggestion
With the release of [React Router v7.2.0](https://github.com/remix-run/react-router/releases/tag/react-router%407.2.0), we’ve introduced a new opt-in framework feature called Split Route Modules. In this post, we’ll explore the performance problem that Split Route Modules solves, how it works, and how to use it today.
```

data/posts/split-route-modules.md
```diff
@@ -0,0 +1,137 @@
...
+  - Mark Dalgleish
+---
+
+With the release of [React Router v7.2.0](https://github.com/remix-run/react-router/releases/tag/react-router%407.2.0), we’ve introduced a new opt-in framework feature called Split Route Modules. In this post we’ll explore the performance problem that Split Route Modules solves, how it works, and how to use it today.
+
+## Route Modules
+
+One of React Router’s defining features in framework mode is the [Route&nbsp;Module&nbsp;API](https://reactrouter.com/start/framework/route-module) which lets you define everything a route needs in a single file. While convenient, this API can sometimes come with a performance tradeoff.
+
+Take for example the following route module:
```
brookslybrand, line 19:
```suggestion
Take, for example, the following route module:
```

data/posts/split-route-modules.md
```diff
@@ -0,0 +1,137 @@
...
+    (response) => response.json()
+  );
+}
+
+export default function Component({ loaderData }) {
+  return <MassiveComponent data={loaderData} />;
+}
+```
+
+Here we have a minimal `clientLoader` that makes a basic fetch call to an external API. The route component, on the other hand, is much larger.
```
brookslybrand, line 35: I would suggest the word "small" or something similar instead of "minimal", just so it is a more obvious antonym to the word "larger".

data/posts/split-route-modules.md
```diff
@@ -0,0 +1,137 @@
...
+
+Unfortunately, because these exports are both contained within the same module, the client loader has to wait for the rest of the route module to download before it can start running. If the client loader needs to perform an async operation like hitting an external API, this delay can be felt by the user.
```
brookslybrand, line 37:
```suggestion
Unfortunately, because these exports are both contained within the same module, the client loader has to wait for the rest of the route module to download before it can start running. If the client loader needs to perform an async operation, like hitting an external API, this delay can be felt by the user.
```

data/posts/split-route-modules.md
```diff
@@ -0,0 +1,137 @@
...
+
+To visualize this as a timeline:
+
+```
+Get Route Module:  |--=======|
+Run clientLoader:            |-----|
+Render:                            |-|
+```
+
+The external API call made from the `clientLoader` is delayed waiting for the hypothetical `MassiveComponent` to download. Since we can't render the component without the data from the client loader, the user has to wait for this network waterfall to complete before the page is rendered.
```
brookslybrand, line 47: Not sure if this is actually better, I just found I had to read this sentence a couple of times trying to keep the 3 pieces of the sentence straight. This was an attempt to reorganize it, but ignore if you feel it's no better

```suggestion
The `clientLoader` is delayed from calling the external API since it has to wait for the hypothetical `MassiveComponent` to download. Since we can't render the component without the data from the client loader, the user has to wait for this network waterfall to complete before the page is rendered.
```

data/posts/split-route-modules.md
```diff
@@ -0,0 +1,137 @@
...
+import { MassiveComponent } from "~/components";
+
+export default function Component({ loaderData }) {
+  return <MassiveComponent data={loaderData} />;
+}
+```
+
+Since these exports have now been split into separate modules, the React Router Vite plugin can ensure that they can be downloaded independently.
+
+This optimization is even more pronounced when more of the Route Module API is used. For example, when using `clientLoader`, `clientAction` and `HydrateFallback`, the timeline for a single route module during a client-side navigation might look like this:
```
brookslybrand, line 88: Trying to avoid the double use of the word "more"

```suggestion
This optimization is even more pronounced when using additional parts of the Route Module API. For example, when using `clientLoader`, `clientAction` and `HydrateFallback`, the timeline for a single route module during a client-side navigation might look like this:
```

data/posts/split-route-modules.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+This would instead be optimized to the following:
+
+```
+Get clientLoader:     |--|
+Get clientAction:     |~~~|
+Get HydrateFallback:  SKIPPED
+Get Component:        |=======|
+Run clientLoader:        |-----|
```
brookslybrand, line 103: I think this is 1 off visually

```suggestion
Run clientLoader:         |----|
```

**before:**
image

**after:**
image

data/posts/split-route-modules.md
```diff
@@ -0,0 +1,137 @@
...
+Render:                        |-|
+```
+
+This looks much better! As before, the client loader doesn't need to wait for the component to download — and now it doesn't need to wait for the `clientAction` or `HydrateFallback` exports to download either. In fact, it doesn't even need to download the `HydrateFallback` export at all during client navigations since it's only ever used on the initial page load.
```
brookslybrand, line 107: I would save your use of `—` for more "interruption" style sentences. Here a comma with the word "and" works perfectly

```suggestion
This looks much better! As before, the client loader doesn't need to wait for the component to download, and now it doesn't need to wait for the `clientAction` or `HydrateFallback` exports to download either. In fact, it doesn't even need to download the `HydrateFallback` export at all during client navigations since it's only ever used on the initial page load.
```

data/posts/split-route-modules.md
```diff
@@ -0,0 +1,137 @@
...
+import type { Config } from "@react-router/dev/config";
+
+export default {
+  future: {
+    unstable_splitRouteModules: "enforce",
+  },
+} satisfies Config;
+```
+
+As always, we'd love to hear your feedback. If you run into any problems or have any suggestions for improvements, please [file an issue](https://github.com/remix-run/react-router/issues) or [start a discussion](https://github.com/remix-run/react-router/discussions).
```
brookslybrand, line 137: Removing the double "any"

```suggestion
As always, we'd love to hear your feedback. If you run into any problems or have suggestions for improvements, please [file an issue](https://github.com/remix-run/react-router/issues) or [start a discussion](https://github.com/remix-run/react-router/discussions).
```

brookslybrand (dismissed):
Great stuff Mark! I mostly went through and added commas, removed double words, and tweaked some sentences. Feel free to ignore any suggestions you don't like.

data/posts/split-route-modules.md
brookslybrand: My 2 cents: I'd prefer us to not wait until the feature is stabilized. Blogs are a good place to talk about features no matter where they are in the process. Docs are where the most up to date information can always been found.

I'm down for a disclaimer saying this is currently an unstable API. We can either update/remove when it goes stable.

Happy to hear Mark's thoughts though

data/posts/split-route-modules.md
markdalgleish: Yeah, I'd love to use the same diagram style as your blog post if we can.

data/posts/split-route-modules.md
markdalgleish: No, the `clientAction` download is just preloading in this case.

data/posts/split-route-modules.md
markdalgleish: Good point, I could expand on this.

data/posts/split-route-modules.md
markdalgleish: Please don't rob me of my em dashes.

data/posts/split-route-modules.md
markdalgleish: No, this is actually correct. Since we're not calling the `clientAction` in this example, we preload it but it doesn't block the running of the `clientLoader`.

markdalgleish: I might add a note clarifying this point.

data/posts/split-route-modules.md
markdalgleish: I've added a note to try and cover this, I'd appreciate your thoughts on it.

data/posts/split-route-modules.md
brophdawg11: DM'd you a link to the excalidraw diagram

data/posts/split-route-modules.md
```diff
@@ -0,0 +1,220 @@
...
+ogImage: /blog-images/headers/react-router-v7.jpg
+imageAlt: "The React Router logo"
+imageDisableOverlay: true
+authors:
+  - Mark Dalgleish
+---
+
+With the release of [React Router v7.2.0](https://github.com/remix-run/react-router/releases/tag/react-router%407.2.0), we’ve introduced a new opt-in framework feature called Split Route Modules. In this post, we’ll explore the performance problem that Split Route Modules solves, how it works, and how to use it today.
+
+Please note that this feature is currently experimental, enabled by the `future.unstable_splitRouteModules` flag. If you want to use it in production, we recommend testing against your production build to ensure that the optimization is working as expected.
```
brophdawg11, line 15: Slight rewording suggestion to better align with the wording in our docs (and add a cross-link):

```suggestion
Please note that this feature is currently [unstable](https://reactrouter.com/community/api-development-strategy#unstable-flags), enabled by the `future.unstable_splitRouteModules` flag. We'd love any interested users to play with it locally and provide feedback, but we do not recommend using it in production yet. If you do choose to adopt this flag in production, please ensure you do sufficient testing against your production build to ensure that the optimization is working as expected.
```

markdalgleish: Nice, I like this.

data/posts/split-route-modules.md
markdalgleish: I've updated the PR with some diagrams. Thanks for the help, it's looking much better now.

PR #346, opened 2025-06-18
Add React Server Components blog post

markdalgleish:


data/authors.yml
```diff
@@ -31,3 +31,6 @@
 - name: Tim Quirino
   avatar: /authors/profile-tim-quirino.png
   title: Staff Product Designer
+- name: Jacob Ebey
+  avatar: /authors/jacob-ebey.jpg
+  title: Senior Developer
```
brookslybrand, line 36: I mean maybe Tim had a bigger impact on RSC than I realized, but I think it should be:

```suggestion
- name: Jacob Ebey
  avatar: /authors/jacob-ebey.jpg
  title: Senior Developer
```

brookslybrand: I'm dumb, this is the `authors.yml` file 🙃

data/posts/react-router-and-react-server-components.md
```diff
@@ -0,0 +1,149 @@
...
+
+This approach has served us well, but it requires significant bundler integration and complexity.
+
+## The New Way: React Router RSC
+
+React Router RSC takes a different approach that's actually simpler:
+
+- Routes defined in app code like any other data structure (no more special `routes.ts` bundler plugins or build-time generation required)
+- A function that handles routing HTML vs data requests
+- A function that generates RSC payloads that get rendered to HTML somewhere
```
brookslybrand, line 115: Trying to match the tense of the bullets above so it's easier to compare and contrast

I will say, the second 2 lines feel a bit hand-wavy. It could just be my unfamiliarity, but I worry that these sentences don't necessarily make it sound "simpler". Are you saying it's simpler because a function is handling this rather than a bundler/file generation?

```suggestion
- You define routes in app code like any other data structure (no more special `routes.ts` bundler plugins or build-time generation required)
- A function handles routing HTML vs data requests
- A function generates RSC payloads that get rendered to HTML somewhere
```

data/posts/react-router-and-react-server-components.md
```diff
@@ -0,0 +1,149 @@
...
+
+Notice that these are now just plain data structures and function calls. These APIs replace what previously required an entire bundler plugin and manifest generation system. The result? A simpler, more direct architecture that allows React Router to be a library again for complex full stack apps, only possible thanks to RSC.
```
brookslybrand, line 117: Wrote the previous comment too soon -- I think this answers it. I wonder if this should preface the bullet points. Take it or leave it, but something like:

```
React Router RSC takes a different approach that's actually simple. Instead of relying on a bundler and plugin manifest system:

- You define routes in app code like any other data structure (no more special `routes.ts` bundler plugins or build-time generation required)
- A function handles routing HTML vs data requests
- A function generates RSC payloads that get rendered to HTML somewhere

Instead of a complex implementation hidden behind a bundler, code splitting is now the result of plain data structures and function calls. The result? A simpler, more direct architecture that allows React Router to be a library again for complex full stack apps, only possible thanks to RSC.
```

That may not be better, so if you don't like it feel free to ignore. I was just a bit confused reading through the first time, so was trying to think through rearranging the text to see if it walked through the argument a little more clearly

data/posts/react-router-and-react-server-components.md
```diff
@@ -0,0 +1,149 @@
...
+
+It's worth calling out that this simplification is useful _even if you're not using server components_. Even if all of your routes are client components, as they are today, this new architecture means that advanced usage of React Router will no longer require a bundler-specific Framework Mode plugin.
+
+## What This Means for You
+
+- **If you're using React Router in Framework Mode:** Our intention is for the future of Framework Mode to be built on top of React Router RSC, but your existing code will continue to work. When the transition happens, it will be seamless from your perspective. _No migration necessary!_
```
brookslybrand, line 123:
```suggestion
- **If you're using React Router in Framework Mode:** Our intention is for the future of Framework Mode to be built on top of React Router RSC, so your existing code will continue to work. When the transition happens, it will be seamless from your perspective. _No migration necessary!_
```

data/posts/react-router-and-react-server-components.md
```diff
@@ -0,0 +1,149 @@
...
+
+- **If you're using React Router in Library Mode:** You can continue using our existing non-RSC library APIs to build more traditional SPAs and SSR apps, both now and into the future.
```
brookslybrand, line 125: 2 things

1. We moved away from "Library mode" and refer to it as ["data" and "declarative" mode](https://reactrouter.com/start/modes). 
2. Might want to speak to the future here: "In a later version you'll easily be able to leverage RSC if you want to without having to adopt framework mode" or something. Not sure that's the exact right way to say it, but I think that's a huge point I don't want to get missed

data/posts/react-router-and-react-server-components.md
```diff
@@ -0,0 +1,149 @@
...
+
+- **If you're starting a new React Router project:** React Router RSC is still unstable, so we currently recommend using the existing Framework Mode or library APIs.
+
+- **If you're feeling adventurous:** We'd encourage you to try our initial unstable RSC APIs and provide feedback, file issues, or even help fix any bugs you run into. We'd love to hear what you think!
+
+### What's Coming
+
+We are planning RSC-enabled Framework Mode for Vite. This depends on some low-level bundler work, but when it lands, Framework Mode becomes much simpler:
+
+1. Create a real routes config from your `routes.ts` convention
```
brookslybrand, line 135: Should we mention anything about backwards compatibility here? I imagine some people may read this and say "they're seriously changing their routing convention _again_!?!?"

brookslybrand (commented):
Really great stuff @jacob-ebey! This article has come a long way and I think explains the benefit quite well

Added a bunch of little comments just with wording suggestions. Take or leave them as you'd like

data/posts/react-router-and-react-server-components.md
```diff
@@ -0,0 +1,149 @@
...
+image: /blog-images/headers/react-router-open-governance.jpg
+ogImage: /blog-images/headers/react-router-open-governance.jpg
+imageAlt: Robot hands passing a disc labeled with the React Router logo
+imageDisableOverlay: true
+authors:
+  - Jacob Ebey
+  - Mark Dalgleish
+---
+
+If you're using React Router today, you're probably wondering what React Server Components (RSC) mean for your applications. The good news? React Router is getting simpler, not more complex. Let's dive into what's happening and what it means for you.
```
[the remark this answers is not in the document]
brophdawg11: I have to read further still but I think it's fine to avoid those types of detail in the lede here if we elaborate on them later on

brophdawg11 (commented):
Looking good!

PR #351, opened 2025-08-20
redirect requests for v2 docs and earlier to v2.remix.run/docs or to …

pcattori:
…GitHub

### DIFF
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -10,6 +10,7 @@ import {
   useMatches,
   useRouteError,
   data,
+  redirect,
 } from "react-router";
 import type { LoaderFunctionArgs } from "react-router";
 import {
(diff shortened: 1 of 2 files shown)


app/root.tsx
```diff
@@ -25,6 +26,69 @@ import iconsHref from "~/icons.svg";
...
+      // So here we use leading `v` in docs ref to detect that older tag format
+      //
+      // See:
+      // - https://remix.run/docs/en/v1.6.4
+      // - https://remix.run/docs/en/1.6.5
+      //
+      // - https://github.com/remix-run/remix/releases/tag/v1.6.4
+      // - https://github.com/remix-run/remix/releases/tag/remix%401.6.5
+
+      // todo: remove leading `v` from newer versions like v1.6.5??
```
pcattori, line 63: For version 1.6.5 and newer, the server redirects things like `vx.y.z` to `x.y.z`. But that should only matter if you manually type that version number into the URL since nothing actually links to `v`-prefixed versions newer than 1.6.5. I assume we just added that at the time to smooth over the 1.6.5 release, but feels like its a non-issue at this point. But let me know what you think!

app/root.tsx
```diff
@@ -25,6 +26,69 @@ import iconsHref from "~/icons.svg";
...
+    }
+
+    // /docs/en/main -> v2.remix.run
+    // /docs/en/dev -> v2.remix.run
+    if (ref === "main" || ref === "dev") {
+      throw redirect(`https://v2.remix.run/docs/${doc}`);
+    }
+
+    // /docs/* -> v2.remix.run
+    throw redirect(`https://v2.remix.run${pathname}`);
```
pcattori, line 85: @brophdawg11 you mentioned wanted this fallback redirect. Right now as it stands, it still will redirect things like `docs/en/blah` to v2.remix.run with the `$lang` still in place. Not sure if that matters since I think we've covered all of our bases for reasonable version numbers and `ref`s in prior checks. I assume this is mostly here to catch any edge-cases we haven't thought about that likely wouldn't contain `en`/`$lang` in the first place

app/root.tsx
brookslybrand: Yeah, I've seen this before in [the react-router-website repo](https://github.com/remix-run/react-router-website/blob/main/app/modules/gh-docs/.server/tags.ts#L87-L91). I bet it's the same reason: with changesets we switched the format of the tags

app/root.tsx
brophdawg11: Could we just strip `en` in those cases?  That should be safe because there's no `docs/en` folder it would otherwise be referring to.  

```
https://remix.run/docs/en/hooks/use-async-error -> https://v2.remix.run/docs/hooks/use-async-error
```

app/root.tsx
```diff
@@ -25,6 +26,69 @@ import iconsHref from "~/icons.svg";
...
+    }
+
+    // /docs/en/0.x -> github
+    // /docs/en/1.x -> github
+    // /docs/en/2.x -> github
+    if (ref.startsWith("0.") || ref.startsWith("1.") || ref.startsWith("2.")) {
+      throw redirect(
+        `https://github.com/remix-run/remix/tree/remix%40${ref}/docs/${doc}`,
+      );
+    }
```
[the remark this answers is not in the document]
brophdawg11: yeah I think this semver detection stuff an be simplified a bit and should just split into a different tag format before `1.6.5` versus after.  Whether or not it has a leading v shoudlnt' make a difference in what we want to do since that's what is currently opting the non-`v` versions less than `1.6.5` into the wrong tag format.

```tsx
let trimmedRef = ref.replace(/^v/, "");
if (["0", "1", "2"].includes(trimmedRef[0])) {
  // We switched from `vx.y.z` tags to `remix@x.y.z` after v1.6.4 release
  let repoTag = versionIs164orLess(trimmedRef)
    ? `v${trimmedRef}`
    : `remix%40${trimmedRef}`;
  throw redirect(
    `https://github.com/remix-run/remix/tree/${repoTag}/docs/${doc}`,
  );
}
```

app/root.tsx
```diff
@@ -25,6 +26,69 @@ import iconsHref from "~/icons.svg";
...
+
+    // /docs/en/main -> v2.remix.run
+    // /docs/en/dev -> v2.remix.run
+    if (ref === "main" || ref === "dev") {
+      throw redirect(`https://v2.remix.run/docs/${doc}`);
+    }
+
+    // /docs/* -> v2.remix.run
+    throw redirect(`https://v2.remix.run${pathname}`);
+  }
```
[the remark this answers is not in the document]
brookslybrand: On that note I bet we can just have a `/docs` route and put this middleware in there, that way it's not running in root

Although eventually we may move some other things over to v2.remix.run, so maybe a big check in here is ideal 🤔

Probably something to do in the cleanup PR anyway if we do that

app/root.tsx
pcattori: Oh yea I've been in the Remix v3 headspace where we are avoiding dependencies where ever possible, but forgot that this repo already has `semver` installed. I'll switch over to that

app/root.tsx
```diff
@@ -25,6 +28,55 @@ import iconsHref from "~/icons.svg";
...
+const redirectOldDocs: Route.unstable_MiddlewareFunction = ({ request }) => {
+  const { pathname } = new URL(request.url);
+
+  if (pathname === "/docs") {
+    throw redirect(`https://v2.remix.run/docs`);
+  }
+
+  if (!pathname.startsWith("/docs/")) {
+    return;
+  }
```
brophdawg11, line 42: Added this back now that we're not automatically constrained to `/docs` URLs based on the route file matching

app/root.tsx
```diff
@@ -25,6 +28,55 @@ import iconsHref from "~/icons.svg";
...
+      throw redirect(`https://v2.remix.run/docs/${path.join("/")}`);
+    }
+
+    // Fallback: /docs/en/* -> v2.remix.run/docs/*
+    throw redirect(`https://v2.remix.run/docs/${ref}/${path.join("/")}`);
+  } else {
+    // Fallback /docs/* -> v2.remix.run/docs/*
+    throw redirect(
+      `https://v2.remix.run/docs/${fullPathWithoutDocs.join("/")}`,
+    );
```
brophdawg11, line 73: This was missing to handle urls like `/docs/foo/bar/baz`

PR #352, opened 2025-08-21
Remove docs

brookslybrand:


app/routes/$.tsx
```diff
@@ -53,18 +52,7 @@ export const loader = async ({ request, params }: Route.LoaderArgs) => {
   handleRedirects(new URL(request.url).pathname);
   handleStaticFileRequests(params["*"]);
```
brookslybrand, line 53: Probably can move this stuff to middleware at some point, but I'd rather do that in another PR

app/routes/$.tsx
```diff
@@ -53,18 +52,7 @@ export const loader = async ({ request, params }: Route.LoaderArgs) => {
...
-    let lang = "en";
-    let doc = await getRepoDoc(ref, `docs/${params["*"]}`);
-    if (!doc) throw null;
-    // FIXME: This results in two fetches, as the loader for the docs page will
-    // repeat the request cycle. This isn't a problem if the doc is in the LRU
-    // cache but we should probably fix it anyway.
-    return redirect(`/docs/${lang}/${ref}/${params["*"]}`);
-    // eslint-disable-next-line no-empty -- should probably do something here
-  } catch {}
-  throw data({}, { status: 404 });
```
brookslybrand, line 67: Shouldn't need any special docs handling in here. Middleware handles most of the cases, and these extra cases we're just not gonna handle

package.json
```diff
@@ -37,17 +36,12 @@
...
-    "follow-redirects": "^1.15.6",
     "front-matter": "^4.0.2",
-    "gray-matter": "^4.0.3",
-    "gunzip-maybe": "^1.4.2",
     "isbot": "^4",
-    "iso-639-1": "^3.1.0",
     "lru-cache": "^10.1.0",
     "luxon": "^3.4.4",
     "morgan": "^1.10.0",
     "octokit": "^3.1.2",
```
brophdawg11, line 44: If this isn't gone I don't think we're totally decoupled from github?

Looks like `lib/resources.server` and `lib/github.server` both use it, and some `routes/_extras.resources` files.  Those should also all be able to be deleted and then we can remove the dep?

brookslybrand: Correct, I meant to call this out so thanks for commenting on it

I'm a bit hesitant to delete them just in case someone is linking. How do you feel about me adding them to the v2 docs as part of the pre-build and setting up some redirects? Or am I being ridiculous and we should just bite the bullet?

brookslybrand: Agreed offline this is safe. Will follow up and move the resources to prerendering on v2 docs for posterity

PR #435, opened 2026-04-08
refactor app runtime around remix/script-server

brookslybrand:
## Summary
- run the app directly from source with `tsx` and route browser module requests through `remix/script-server` instead of Vite's app runtime
- switch document and hydrated entries to `/scripts/**` resolution with `clientEntry(import.meta.url, ...)`, `loadScriptEntry()`, and SSR `resolveClientEntry`
- clean up shared browser-safe modules so client entries reuse canonical UI modules without duplicating components across `app/assets` and `app/ui`

## Testing
- `pnpm run typecheck`
- `pnpm test -- --run`
- `pnpm run build`


app/utils/assets.server.ts
```diff
@@ -0,0 +1,25 @@
...
+  scripts: {
+    sourceMaps: isDevelopment ? "external" : undefined,
+    minify: !isDevelopment,
+    define: {
+      "process.env.NODE_ENV": JSON.stringify(
+        process.env.NODE_ENV ?? "development",
+      ),
+    },
+  },
+  watch: isDevelopment,
```
markdalgleish: If you're using `tsx watch` to reboot the server on file changes, you should just always disable watch.

PR #444, opened 2026-05-01
Polish landing page mobile layout

tquirino:
## Summary

A pass of small CSS adjustments to the landing page, focused on the mobile (`<= 880px`) viewport.

### Hero (`landing-hero.tsx`)

- Drop the `text-align: justify` + `text-align-last: left` + `hyphens: auto` fallback applied at `<= 880px` and replace with plain `text-align: left`. The hero paragraph now reads left-aligned to match the surrounding text panel rather than producing the stretched word-spacing of justified text on narrow lines.

### Mobile nav (`landing-nav.tsx`)

- Add `margin-top: -12px` and `margin-right: -12px` to the mobile container so the MENU/CLOSE button optically aligns with the Remix wordmark on the opposite side of the header (the wordmark is 16px tall, anchored at `top: 24px`; the button was 40px tall, anchored at `top: 24px`, so its label center sat 12px below the wordmark center).
- Swap the "close" text label for an inline `<svg>` X glyph at `width: 1em; height: 1em` with `stroke="currentColor"`, so it inherits the button's font-size and color exactly. Marked `aria-hidden` since the button's `aria-label="Close menu"` already conveys meaning.
- Right-align the open menu's links via `justify-content: flex-end` on `mobileMenuItemStyles` (previously inheriting `center` from `navItemStyles`).
- Tighten the popover anchor offset from `8` to `4` so the gap between the toggle and the menu matches the `4px` `gap` between menu items.

### Package logos (`package-logos.tsx`)

- Introduce `MOBILE_LOGO_VIEWPORT_GUTTER_PX = 40` and a unified `gutterPx` derived at runtime from the measured `panelLeft`, so the logo group sits exactly 40px from each viewport edge on mobile while continuing to use the desktop `VIEWPORT_GUTTER_PX = 24` everywhere else. The Data logo's offset and the per-logo `right` value all share the same `gutterPx` so the composition shifts inward together.
- Precompute `WIDEST_LOGO_WIDTH_FRACTION` (the widest logo's width as a fraction of container height) and use it to scale `stackedHeight` on mobile so the widest single logo spans `panelWidth - 2 * gutterPx`. The existing scattered desktop layout is preserved.
- Bump Auth's `heightFraction` from `0.1786` to `0.242` so its rendered width matches Routing's (Auth's narrower aspect ratio previously left it noticeably shorter).
- Dynamically compute `dataSessionHeightFraction` so `dataWidth + 30px gap + sessionWidth` exactly equals the widest logo's width at any container height. The `Math.max` floor at the original `0.1607` guarantees we never shrink the row below its previous size. This applies on both desktop and mobile so Data + Session always line up vertically with the rows above and below.
- Nudge the Component logo up by 8px (`top: calc(75% - 8px)`) to tighten its spacing under the Data + Session row.

### Footer (`landing-footer.tsx`)

- Add a `MOBILE_BREAKPOINT_PX = 880` and at that breakpoint flip the brand row, social icon row, legal column, and outer container from `flex-end` to `center` (with `text-align: center` on the legal block to handle wrapped lines). Desktop alignment is unchanged.

## Test plan

- [x] At a mobile viewport (`<= 880px wide`):
  - [x] Hero paragraph is left-aligned, no stretched word spacing.
  - [x] Header MENU button's label sits on the same horizontal line as the Remix wordmark on the left, with a 12px top gap and a 12px right gap.
  - [x] Opening the menu shows an X icon (not the word "CLOSE"), same color and visual height as the surrounding labels.
  - [x] Menu links right-align inside their buttons; the gap between the X button and the first menu item matches the gap between menu items.
  - [x] Package logo group is inset 40px from each viewport edge; Auth and Routing share the same width; Data + Session combined width matches the widest logo above and below; Component logo sits 8px closer to Data/Session than before.
  - [x] Footer brand row, social icons, and legal text are horizontally centered.
- [x] At a desktop viewport (`> 880px wide`):
  - [x] Hero, nav, and footer are visually unchanged.
  - [x] Package logos: scattered absolute layout is unchanged except Auth is now wider (matching Routing) and Data + Session combined width lines up with the rows above and below.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/app/assets/remix-landing/components/landing-footer.tsx
+++ b/app/assets/remix-landing/components/landing-footer.tsx
@@ -26,6 +26,8 @@ const SOCIAL_LINKS = [
   },
 ] as const;
 
+const MOBILE_BREAKPOINT_PX = 880;
+
 const footerShellStyles = css({
   padding: "40px 24px 24px",
 });
@@ -36,6 +38,9 @@ const footerContentStyles = css({
   alignItems: "flex-end",
   justifyContent: "flex-end",
   gap: "12px",
+  [`@media (max-width: ${MOBILE_BREAKPOINT_PX}px)`]: {
+    alignItems: "center",
+  },
 });
 
 const brandRowStyles = css({
@@ -45,6 +50,9 @@ const brandRowStyles = css({
   gap: "24px",
   width: "100%",
   flexWrap: "wrap",
+  [`@media (max-width: ${MOBILE_BREAKPOINT_PX}px)`]: {
+    justifyContent: "center",
+  },
 });
 
 const wordmarkStyles = css({
@@ -58,6 +66,9 @@ const socialLinksStyles = css({
   alignItems: "center",
   justifyContent: "flex-end",
   gap: "16px",
+  [`@media (max-width: ${MOBILE_BREAKPOINT_PX}px)`]: {
+    justifyContent: "center",
+  },
 });
 
 const socialLinkStyles = css({
@@ -91,6 +102,10 @@ const legalStyles = css({
   textTransform: "uppercase",
   color: colors.fg,
   whiteSpace: "nowrap",
+  [`@media (max-width: ${MOBILE_BREAKPOINT_PX}px)`]: {
+    alignItems: "center",
+    textAlign: "center",
+  },
 });
 
 const legalTextStyles = css({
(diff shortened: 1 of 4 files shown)


brookslybrand (approved):
Looks great! Swapped for the existing X icon

PR #446, opened 2026-05-01
Better webgl (opinionated)

vallafederico:
Mostly personal preference but close to alll the perf we recovered was invested in a nicer GPGPU mouse interacion. Also some states orchestration, baking of non time related states and some overall small improvements.


brookslybrand:
Thanks @vallafederico for all these changes! Just reviewed with our designer @tquirino 

> omg. YES
> ship it all
> these are all definite improvements and I didn't notice any regression in perf

The one I did take out is the copy interaction. I tried adding this initially, but I want to think about it a little more. I would like to add it, just want to think a bit more about how the interaction feels and add it in a separate PR, not lump it in with all these perf improvements

Again, thank you so much for taking time to tackle these issues and improve the performance and UX so much!

vallafederico:
my pleasure, love you guys and the work you're doing <3

PR #449, opened 2026-05-04
Add "A Brand New Remix" blog post and inline image lightbox

tquirino:
## Summary

- Adds a new **draft** blog post, "A Brand New Remix" (`data/posts/brand-new.md`), with header and inline images covering the brand evolution and the new website's implementation.
- Adds a `BlogLightbox` client entry (`app/assets/blog-lightbox.tsx`) that decorates inline images inside `.md-prose` and opens them in a fullscreen overlay. Includes focus management, scroll lock, click/Enter/Space to open, click-backdrop or Escape to close.
- Aligns inline blog image width with the header image on `lg+` viewports by matching the `> p:has(img)` negative margin (`-mx-12` → `-mx-10`) to the `container`'s configured `lg` padding (`2.5rem`). Previously inline images overflowed the prose container by 16px on each side at `lg+`.

## Implementation notes

- `BlogLightbox` is mounted as a sibling of `<main>` (under `<Document>`), not as a child of `<main>`. Placing it inside `<main>` as a sibling of the prose `<div innerHTML={post.html}>` caused the Remix UI runtime to non-deterministically crash during hydration in `collectFrameServerStyleTags` because the prose `innerHTML` subtree isn't represented in the JSX VDOM that the runtime walks. Mounting the lightbox as a top-level hydration island keeps its diff fully independent of the prose container, which is also the conventional placement for a `position: fixed` overlay.
- The lightbox uses document-level event delegation to detect clicks on `.md-prose img` and avoids hijacking images that are wrapped in anchor tags.

## Test plan

- [ ] Visit `/blog/brand-new` (the post is `draft: true`, so it won't appear in listings — navigate directly).
- [ ] Confirm the header image and all four inline images render at the same width on `lg+` viewports.
- [ ] Click each inline image and confirm the lightbox opens fullscreen, scaling up to a max of 1600px or the viewport width.
- [ ] With the lightbox open, confirm focus moves to the close button.
- [ ] Press `Escape` and confirm the lightbox closes and focus returns to the originally-clicked image.
- [ ] Click the backdrop (outside the image) and confirm it also closes the overlay.
- [ ] Activate an image with the keyboard (Tab to focus, Enter or Space to open).
- [ ] Visit an existing blog post (e.g. `/blog/wake-up-remix`) and confirm prose rendering, links, and code blocks are unaffected.
- [ ] In the browser devtools console, confirm there is no `TypeError: Cannot read properties of null (reading 'childNodes')` on initial page load.


data/posts/brand-new.md
```diff
@@ -0,0 +1,64 @@
...
+imageDisableOverlay: true
+---
+
+Brands typically change to signal a shift in thinking or reflect a changing audience. Sometimes, it happens as a way to adapt to a different landscape. Some may see this as a way to align with current trends, but with a name like Remix, is it really that surprising that we would want to reinvent ourselves? Everything is rapidly evolving around us, and my goal as the designer on this team was to craft a new identity that would allow us to drive into the future.
+
+## Context & Trigger
+
+![Brand evolution board mapping Remix's Past, Present, and Future across three rows of context cards.](/blog-images/posts/brand-new/frames-brand-thinking.avif)
+
+Ultimately, the trigger — the moment we decided we needed a new brand — was when we established [our philosophy for building Remix 3](https://github.com/remix-run/remix/blob/main/README.md) and beyond.
```
brookslybrand: I think we should link to the blog post instead of GitHub

```suggestion
Ultimately, the trigger — the moment we decided we needed a new brand — was when we established [our philosophy for building Remix 3](/blog/wake-up-remix) and beyond.
```

data/posts/brand-new.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Evolving the Logo
+
+![Logo evolution board grouped into three rows: Evolving the Logo, Preserving the DNA, and Logo in use, showing wordmark explorations and applications.](/blog-images/posts/brand-new/frames-logo-evolution.avif)
+
+The new identity was meant to have a clear lineage from its predecessor, and it came quite naturally as we were putting together the [demo for last year's Remix Jam](https://remix.run/jam/2025/gallery?photo=46). Customizing letterforms from the Kanada typeface also seemed like a nice reference to our parent company's Canadian roots. Speaking of which, this was the first substantial shift in the brand since Remix joined Shopify, so I felt that it was necessary for the two brands to work together in harmony. Our racing theme already nudged the logo and wordmark into an italicized direction, so this worked in our favor.
```
brookslybrand: Worth linking to Kanada typeface?

data/posts/brand-new.md
tquirino: Yep, sounds like a good change!

PR #465, opened 2026-06-04
Add Remix docs routes

mjackson:
## Summary
- Add the `/docs` route map, docs controller, index page, and 16 placeholder chapter routes.
- Render docs pages through the request context `render` function and pass the current request URL into social head tags.
- Point the header Docs link at the new local docs route and verify `/docs` is served before legacy catchall redirects.

## Verification
- `pnpm exec prettier --write` on changed files
- `pnpm exec remix test app/controllers/docs/controller.test.ts app/controllers/catchall.test.ts`
- `pnpm exec tsc --noEmit`
- `pnpm run build`


mjackson:
Closing — the Remix 3 docs are moving to the main `remix-run/remix` repo instead of living in the website.

They're now a standalone, runnable Remix app under `guides/` in the monorepo (route map, docs controller, index page, shared layout, and the 16 placeholder chapters), so we can run the app and browse the docs as we flesh them out. The chapter/index source was ported over as-is; only the website-specific chrome (header/footer, social tags, Tailwind build) was swapped for a small standalone shell + stylesheet.

Not merging this here.

PR #466, opened 2026-06-04
Refactor docs routes to use markdown content

brookslybrand:
## Summary
- move docs chapter content into markdown files under data/docs
- collapse hardcoded chapter routes/controllers into a markdown-backed docs controller
- preserve the existing docs page appearance with dedicated docs CSS

Stacked on #465.

## Verification
- pnpm exec remix test app/controllers/docs/controller.test.ts
- pnpm run build
- pnpm exec oxlint . --threads=1


brookslybrand:
Micheal prefers JSX for now. Gonna revisit this at a latter point, maybe with some MDX-like experience

My thoughts:
- I definitely want the ability to put some rich components in here and interactive experiences
- I find markdown much easier to contribute to, parse, read, re-use, feed to agents, etc.

So I'd love to move more toward markdown and away from this duplicated experience. As we get further along with our content a solution should be easier to find.

PR #476, opened 2026-07-23
Preload SSR-discovered client entries

brookslybrand:
## Why

SSR already discovers every rendered `clientEntry`, but the document previously exposed those modules only through Remix's serialized hydration data. The bootstrap entry had to load before the browser could discover route hydration roots, producing avoidable module waterfalls. Home-specific preload lists reduced that waterfall for one route while duplicating information Remix already had.

Calling `assetServer.getPreloads()` for each discovered root was too aggressive: it traverses literal dynamic imports too, eagerly fetching intentionally lazy features such as the FPS counter and Jam's capability-gated volumetric renderer.

## How

- Collect the public href of each `clientEntry` resolved during SSR and emit one `modulepreload` per discovered hydration root.
- Continue using `getPreloads()` for the bootstrap entry's known graph, but do not recursively expand discovered hydration roots. Dynamic imports therefore remain lazy.
- Hold only Remix's initial render chunk long enough to insert preload links before the bootstrap script; subsequent document and frame chunks continue streaming unchanged.
- Emit frame-local client-entry preloads with frame content, then deduplicate entries shared by a frame and its parent document.
- Remove home-specific browser-entry lists and preload mutation from route controllers and the document component.
- Statically import the immediately used particle canvas while retaining event- and capability-triggered dynamic imports.
- Use Remix's normal client-entry export resolution instead of the application-level fallback.

The render and frame architecture follows the conventional middleware proposed in [remix-run/remix#11607](https://github.com/remix-run/remix/pull/11607): centralized asset resolution, top-frame propagation, safe same-origin header forwarding, strict cross-origin header allowlisting, redirect-body cancellation, and support for source, public, and CDN client-entry URLs. We retain `renderWith()` only for preload injection and the Jam-specific SSR frame marker; this should make a future move to `render({ assets })` straightforward once the standard middleware exposes preload integration.

## Result

- Home, blog, Jam, and future routes automatically preload SSR-discovered hydration roots.
- Dynamic children such as `fps-counter.tsx` and the volumetric cloud renderer are not explicitly preloaded.
- Bootstrap, document, and nested-frame preload links are deduplicated.
- Typed route hrefs remain inside browser components; no asset-only href props are required.

## Validation

- `pnpm run build`
- `pnpm test` — 144 passed
- targeted oxlint on changed files
- Prettier and `git diff --check`
- production fingerprint rendering for home, blog, and Jam preload graphs

### DIFF
--- a/app/assets/entry.ts
+++ b/app/assets/entry.ts
@@ -6,22 +6,7 @@ initFathomAnalytics();
 let app = run({
   async loadModule(src, exportName) {
     let mod = await import(src);
-
-    let exp = (mod as Record<string, unknown>)[exportName];
-    if (typeof exp === "function") return exp;
-
-    // Minified builds may rename exports (e.g. NewsletterSubscribeForm -> N).
-    // Fallback: find a function with clientEntry metadata.
-    for (let value of Object.values(mod as object)) {
-      if (
-        typeof value === "function" &&
-        (value as { $entry?: boolean }).$entry === true
-      ) {
-        return value;
-      }
-    }
-
-    throw new Error(`Export "${exportName}" from "${src}" is not a function`);
+    return mod[exportName];
   },
   async resolveFrame(src, signal, target) {
     let headers = new Headers();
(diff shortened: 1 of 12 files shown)


brookslybrand:
We still do have a waterfall, as seen on staging: https://remixdotrunstage.fly.dev/blog, so this is gonna be an issue for other applications as well.

I don't feel I have a good enough grasp of what's going on here to propose a solution.


image

PR #484, opened 2026-08-10
Add HMR script

markdalgleish:
This adds a `pnpm hmr` script to start the app in development mode with HMR enabled. I've updated AGENTS.md so that we use `pnpm dev` by default and only switch to `pnpm hmr` during heavy UI iteration.

There was an issue with the loading screen re-appearing on file changes because of the way it modified the document shell, so to fix this I've moved the state into a hydrated component.

Also, top frame reloads ran into an issue with our production caching headers where it would retain the cached HTML in memory rather than getting fresh updated HTML from the server, so I've made our render middleware make an exception for this in development.

Note, a lot of the diff in this PR in terms of line count is because the updated `pnpm-lock.yml` is formatted differently. I'm running the same version of pnpm declared in our package.json so I'm not sure why it differed - seems likely to me that the lockfile that's checked in was generated with a different version of pnpm.


brookslybrand (approved):
Amazing! Thank you

PR #490, opened 2026-08-19
Optimize blog images with Remix assets

brookslybrand:
## Summary

- serve blog images through Remix's asset transformation system
- generate constrained responsive WebP variants with intrinsic dimensions and production fingerprinting
- optimize blog listing, post hero, Markdown, and trusted raw HTML images while preserving full-size lightbox sources
- add filesystem transform caching and focused route, Markdown, asset, and browser coverage

## Validation

- `pnpm test` (156 tests passed)
- `pnpm run build`
- `pnpm remix doctor`
- `pnpm run typecheck`
- `pnpm exec oxlint . --threads=1`
- production asset route smoke test (`200 image/webp`, immutable one-year cache)

### DIFF
--- a/app/utils/blog-image-assets.ts
+++ b/app/utils/blog-image-assets.ts
@@ -4,55 +4,77 @@ import sharp, { type Metadata } from "sharp";
 import { assets } from "./assets.ts";
 
 const ROOT_DIR = path.resolve(import.meta.dirname, "../..");
+const AVATAR_WIDTHS = [64, 128] as const;
 const RESPONSIVE_WIDTHS = [480, 768, 1200, 1600] as const;
 const TRANSFORMABLE_EXTENSIONS = new Set([".jpeg", ".jpg", ".png"]);
 
 type WebpTransform =
   | "webp"
+  | "webp-64"
+  | "webp-128"
   | "webp-480"
   | "webp-768"
   | "webp-1200"
   | "webp-1600";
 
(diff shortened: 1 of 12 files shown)


app/utils/blog-image-assets.ts
```diff
@@ -0,0 +1,140 @@
...
+  let outputWidths = responsiveWidths.filter((width) => width < sourceWidth);
+  if (sourceWidth <= responsiveWidths.at(-1)!) {
+    outputWidths.push(sourceWidth);
+  }
+
+  let variants = await Promise.all(
+    outputWidths.map(async (width) => ({
+      href: await assets.getHref(filePath, {
+        transform: [getWebpTransform(width, sourceWidth)],
+      }),
```
markdalgleish: Should we have some error handling and a fallback image here? If there are any issues with resolving/processing/transforming the image, it looks like it'll bubble out into breaking the page. I'm not sure of the best way to handle this, but it's something we might want to think about.

brookslybrand: Yep, we definitely should. I just made it to where it falls back to the original image. Definitely don't want a failed transform to take down a page

I thought about doing a 404, but I don't think that's the right UX. If the image itself doesn't exist, then it'll still end up being a 404 since the original image won't show up. A transform though should have a graceful degradation in this case.

PR #496, opened 2026-08-25
Redesign home page content

markdalgleish:
This is my initial draft pass at redesigning the content on the home page to try and give a more detailed overview of what Remix is. This is definitely still a work in progress, but I'm sharing early to get some feedback.

### DIFF
--- a/app/actions/home-landing-content.tsx
+++ b/app/actions/home-landing-content.tsx
@@ -55,7 +55,7 @@ const differentiatorInlineCodeStyles = css({
 const differentiators = [
   {
     title: "State is just JavaScript",
-    body: "A component runs setup once, then returns a function that renders JSX. Keep state in ordinary JavaScript variables, objects, or classes—without hooks or a prescribed state container.",
+    body: "A component runs setup once, then returns a function that renders JSX. Keep state in ordinary JavaScript variables, objects, or classes rather than hooks or a prescribed state container.",
   },
   {
     title: "Updates are explicit",
@@ -64,18 +64,17 @@ const differentiators = [
         Your code decides when the UI renders. Call{" "}
         <code mix={[differentiatorInlineCodeStyles]}>handle.update()</code>{" "}
         after changing state, and await it when your next step depends on the
-        updated DOM. The path from update to screen stays visible and
-        predictable.
+        updated DOM.
       </>
     ),
   },
   {
     title: "Not everything needs a component",
-    body: "Mixins attach reusable behavior directly to individual elements—events, styles, refs, and accessibility—without changing the markup or introducing another component.",
+    body: "Mixins attach reusable behavior di events, styles, refs, and accessibility behavior directly to individual elements. This keeps the markup intact without introducing another component.",
   },
   {
-    title: "Client components, without the bundler magic",
-    body: "Your server runtime decides how hydrated client components map to browser code—without a bundler hiding the boundary from you.",
+    title: "Client components with visible boundaries",
+    body: "You define how hydrated client components map to browser code in the server runtime. The client boundary stays visible in your code.",
   },
   {
     title: "HTML over the wire",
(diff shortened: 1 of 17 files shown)


brookslybrand:
Thanks @markdalgleish! I'm sure styles are very WIP, but just wanted to point out that it's doing some pretty obvious Codex-isms and getting rid of are pretty decent styles

For now, I'm going to focus more on copy and any restructuring going on

image

app/actions/public/remix-landing/components/landing-hero.tsx
```diff
@@ -47,35 +45,99 @@ const headingStyles = css({
...
+const scrollActionIconStyles = css({
+  width: "20px",
+  height: "20px",
+});
+
 export function LandingHero(_handle: Handle) {
   return () => (
     <section id="the-framework" mix={[shellStyles]}>
       <div mix={[textGroupStyles]}>
-        <h1 mix={[headingStyles]}>A web framework for building anything</h1>
```
brookslybrand, line 73: I can't explain to you how much I like this description while also finding it extremely funny. I just keep thinking of this meme:

Image

app/actions/public/remix-landing/components/landing-hero.tsx
```diff
@@ -47,35 +45,99 @@ const headingStyles = css({
...
+        <h1 mix={[headingStyles]}>The fully-stacked web framework</h1>
         <p mix={[bodyStyles]}>
-          Remix gives you the power and tools to build anything you can dream
```
brookslybrand, line 75: This is good, we need something about the UI piece I think, or just need to expand it a bit. Potentially need to lean into the fact that we are shipping components. I see this sentiment a lot:

https://x.com/tsAngelNunnez/status/2092060594547413420

brookslybrand:
I think this is the right move, similar to how what I have on the [quick site](https://remix.quick.shopify.io/#:~:text=A%20framework%20organized%20by%20purpose). I think there's room here to some how really punch home the point that it's all 1 `remix` package. That means cohesive ecosystem, less bloat (because of the modular design), and more secure by default (less attack vectors). Might not be the right place to include it, and you go into some of that later, but just wanted to throw that in.

image


image

app/actions/home.tsx
```diff
@@ -11,7 +11,7 @@ type HomePageProps = {
   previewImage: string;
 };
 
-const HOME_TITLE = "Remix - A Web Framework for Building Anything";
+const HOME_TITLE = "Remix - The Fully-Stacked Web Framework";
 const HOME_DESCRIPTION =
   "Remix is a batteries-included, ultra-productive, zero dependencies and bundler-free framework, ready to develop with in a agent-first world.";
```
mjackson:
```suggestion
  "Remix is a full-stack web framework with built-in server, data, UI, asset, and testing tools. Build from a single package with first-class support for coding agents.";
```

app/actions/home-landing-content.tsx
```diff
@@ -1,99 +1,160 @@
...
+        title: "Built to be understood",
+        body: "Trace behavior through ordinary code and web standards instead of hidden framework machinery.",
+      },
+      {
+        title: "Built to be changed",
+        body: "Follow the defaults, replace a layer, or take control of the logic when your app needs it.",
+      },
+      {
+        title: "Built to be generated",
+        body: "Remix skills help agents generate code using the framework’s APIs, conventions, and workflows.",
```
mjackson: i think "built for coding agents" is a little more direct.

```suggestion
        title: "Built for coding agents",
        body: "Remix skills teach agents the framework’s APIs, conventions, and workflows.",
```

app/actions/home-landing-content.tsx
```diff
@@ -1,99 +1,160 @@
...
+    title: "Rendering is explicit",
+    body: (
+      <>
+        Changing state doesn&apos;t trigger rendering. Store it however you
+        like, then call{" "}
+        <code mix={[differentiatorInlineCodeStyles]}>handle.update()</code> when
+        the UI should change. You decide when state reaches the screen.
+      </>
+    ),
   },
```
mjackson, line 70: Can we combine these?

```suggestion
  {
    title: "State and rendering stay explicit",
    body: (
      <>
        Store state in ordinary JavaScript variables, then call{" "}
        <code mix={[differentiatorInlineCodeStyles]}>update()</code> when
        the UI should render. No hooks or reactive runtime.
      </>
    ),
  },
```

brookslybrand: The only downside of combining these two is that it breaks our 2x4 grid

image

app/actions/home-landing-content.tsx
brookslybrand: I feel like the third point essentially just being "we have an Agent Skill" is not that great of a point. Everyone has an Agent Skill. I almost would rather not have this section entirely than leave it as is. I do like having 3 points, so I'll keep thinking of a good 3rd after
1. built to be understood
2. built to be changed
3. ???

PR #502, opened 2026-09-01
Improve hero body contrast on short viewports

brookslybrand:
## Summary
The landing hero paragraph sits on the particle field, and at shorter viewport heights the road overlaps the copy.

- Default: dark text-shadow on the existing gray body color (no layout change)
- `@media (max-height: 980px)`: white type + a stronger halo, matching the sizes where the road hits the text

Heading is unchanged.

## Test plan
- [ ] Tall desktop: paragraph should look close to today, with a slightly darker halo instead of the white glow
- [ ] Short / laptop-height window where the road crosses the copy: body should read more clearly (white + stronger shadow)
- [ ] Resize across ~980px height and confirm the snap

### DIFF
--- a/app/actions/public/remix-landing/components/landing-hero.tsx
+++ b/app/actions/public/remix-landing/components/landing-hero.tsx
@@ -49,6 +49,11 @@ const bodyClosingPhraseStyles = css({
   whiteSpace: "nowrap",
 });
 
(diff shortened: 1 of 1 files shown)


brookslybrand:
Closing — we're not going with the height-based text-shadow.

PR #508, opened 2026-09-17
Add interactive framework overview to home page

markdalgleish:
This reworks the opening section of the home page to show more concrete, code-oriented details about Remix up front.

It adds an interactive overview of the server, data, auth, assets, UI, animation, and primitives APIs, with basic examples that show how the pieces fit together without turning the page into full documentation. Visual previews are included where they help explain the feature, while the rest of the layout focuses on code.


brookslybrand:
We probably want to think through our mobile story a bit. I don't think it's obvious that you scroll through these, and wonder if some sort of select might be better 🤷‍♂️

image

